PR #7, opened 2015-04-17
Print out the fucking command before running it

fordhurley:
This makes thefuck act similar to fixing commands with either `!!` or `fc`, and I think helps to remove some of the "magic". For example, I use `sudo !!` all day long when on I'm linux systems:

```
$ apt-get install something
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
$ sudo !!
sudo apt-get install something
[sudo] password for test:
```


nvbn:
Cool, thank you!

PR #8, opened 2015-04-17
Adds rule for cd.. and .gitignore for vim tmp files

LaMi-:
I guess cd.. is a very common typo which can easily be handled using the attached fuck rule.


nvbn:
Thanks!

PR #9, opened 2015-04-17
Fixed UnicodeDecodeError

alexnastetsky:
Was getting this error:
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 20: ordinal not in range(128)
```
This pull request fixes it.


nvbn:
Thanks!

Issue #1, opened 2015-04-17
running "fuck" twice almost gave me a heartattack


eridal:
```
who@where:~$ fuck
No fuck given
who@where:~$ fuck
fsck from util-linux 2.20.1
e2fsck 1.42.9 (4-Feb-2014)
/dev/sda6 is mounted.

WARNING!!!  The filesystem is mounted.   If you continue you ***WILL***
cause ***SEVERE*** filesystem damage.

Do you really want to continue<n>? no
check aborted.
```

jbrooksuk:
This doesn't happen for me on OSX 10.10, that's interesting.

nvbn:
It's a bit strange, do you wrote something in another terminal between two fucks?

curtisz:
Haha this made my day. However, I could not replicate this on Linux Mint 17 (Ubuntu 14.04).

brianshumate:
The `fsck` would not succeed as a non-root user anyway.

LaMi-:
Can reproduce this on ubuntu 14.10. For me I just need to execute this twice:

```> /usr/lib/command-not-found fuck
No command 'fuck' found, did you mean:
 Command 'suck' from package 'suck' (universe)
 Command 'fsck' from package 'util-linux' (main)
fuck: command not found```

The order of suck and fsck is changed on each call. When fsck is shown first, the error appears.

When I remove the subshell from the shell alias this is easily visible:

```
> fuck
No fuck given
> fuck
fsck
> fuck
suck
```

eridal:
At first, I tested the `apt-get`, as shown in the README, and then tried running fuck again, and that was the first time the error happened.

Then I closed the console, and opened another again; and that's the output you're seeing.

btw, I'm running `Linux where 3.13.0-49-generic #81-Ubuntu SMP Tue Mar 24 19:29:48 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux`

LaMi-:
btw. I just added an additional alias for testing. It simply outputs the command fuck would execute:

```alias tryfuck='thefuck $(fc -ln -1)'```

mvirkkunen:
@brianshumate Wouldn't a third fuck fix that though?

nvbn:
I added check if previous command is `fuck`. It should fix this issue.

brianshumate:
@mvirkkunen given enough fucks, i think this could totally fuck one's system up, yes.

LaMi-:
[...]

PR #10, opened 2015-04-17
Untested: Added --dry as an option.

jacob9706:
This will print the command that will be ran and prompt for yes or no if a command is found.
Example:
`fuck --dry`

### DIFF
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -95,9 +95,12 @@ def main():
         rules = get_rules(user_dir, settings)
         matched_rule = get_matched_rule(command, rules, settings)
         if matched_rule:
+            print(rule.get_new_command(command, settings))
+            
             if opts.do_it_dry:
-                print('Fuck again to run: {0}'.format(rule.get_new_command(command, settings)))
-            else:
-                run_rule(matched_rule, command, settings)
+                answer = input('Run command? [Y/n]: ')
+                if answer.lower() != 'y':
+                    return
+            run_rule(matched_rule, command, settings)
         else:
             print('echo No fuck given')
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -82,8 +82,8 @@ def is_second_run(command):
 
 def main():
     parser = argparse.ArgumentParser(description='The fuck?')
-    parser.add_argument('--dry', dest='do_it_dry', type=bool, default=False,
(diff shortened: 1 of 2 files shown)


jamescostian:
:+1: for this solution that prevents heart attacks (#1)

I don't always fuck, but when I do, I fuck --dry

ahmetalpbalkan:
Dry fuck doesn't sound like a good idea at all. Kills the point of the project.

How about printing the command that's going to be executed, sleeping 1.5 or 2 seconds to give a chance to ctrl+c? `git` autocorrect config does this way and works for most. https://coderwall.com/p/gvw46q/git-autocorrect-spelling

hmemcpy:
Is this like dry humping then?

thefuck/main.py
```diff
@@ -90,6 +95,12 @@ def main():
         rules = get_rules(user_dir, settings)
         matched_rule = get_matched_rule(command, rules, settings)
         if matched_rule:
+            print(rule.get_new_command(command, settings))
+            
+            if opts.do_it_dry:
+                answer = input('Run command? [Y/n]: ')
```
kilotaras: Can empty answer be treated as yes here?

foofoo112:
Yyy

SkyRocknRoll:
I Would like to Have this 'Run command? [Y/n] 

+1

nwinkler:
If a question is asked, it doesn't sound like _dry_ mode to me. With other tools, _dry_ mode doesn't do anything.

The above ("Run command? [Y/n]") looks more like interactive mode to me, so a `-i` instead of `-dry` might be more appropriate.

SkyRocknRoll:
if i want to use this for serious command which modifies or deletes my system data then it would be nice to have confirmation.

nvbn:
I added settings option `require_confirmation`, I think it's what you want, when It's `True`:
```bash
➜  apt-get install vim    
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
➜  fuck
sudo apt-get install vim [Enter/Ctrl+C]
[sudo] password for nvbn: 
```
Or:
```bash
➜  fuck
sudo apt-get install vim [Enter/Ctrl+C]^CAborted
```

PR #19, opened 2015-04-18
Improved installation instructions - easy_install instead of pip

revmischa:
`sudo pip install` fucks everything up and depends on `pip`. You want `easy_install`. Trust me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ git branch
 Install `The Fuck`:
 
 ```bash
-sudo pip install thefuck
+sudo easy_install thefuck
 ```
 
 And add to `.bashrc` or `.zshrc`:


tpltnt:
I advise against merging this pull request based on [SO: Why use pip over easy_install?](http://stackoverflow.com/questions/3220404/why-use-pip-over-easy-install#3220572)

drvinceknight:
+1 @tpltnt

davoclavo:
@revmischa You should add both, at least pip should be included, because in [Python 3.4 pip is included by default](https://docs.python.org/3/installing/#key-terms).

nvbn:
`easy_install` added in the readme as a fallback if `pip` fails.

revmischa:
Installing pypi modules system-wide via `pip` fucks your shit up on OSX. You get all kinds of permissions errors and general fuckery. `pip` should really only be used for a venv.

PR #36, opened 2015-04-18
Improve no_command

jjmason:
This PR fixes a couple of limitations of the no_command rule:

`/usr/lib/command-not-found` suggests commands from uninstalled packages, so we filter out commands that aren't installed. 

In many cases taking the first installed suggestion is inappropriate -- for example, if the broken command is `bish`, on my system two of the suggestions are installed: `wish` and `bash`.  Which one did I mean?  We can guess by counting the number of occurrences in my shell history:
```sh
➜  thefuck git:(improve-no-command) history | egrep '\bbash\b' | wc -l
35
➜  thefuck git:(improve-no-command) history | egrep '\bwish\b' | wc -l
3
```
which tells us that I probably want `bash` (I don't even know what `wish` is, it's just in my history because I `which`ed it and `grep`ed for it a couple of times).

I've updated the tests to check these cases and continue to work with the original cases.

### DIFF
--- a/tests/rules/test_no_command.py
+++ b/tests/rules/test_no_command.py
@@ -50,8 +50,7 @@ def bin_might_exist(request):
 @pytest.fixture
 def patch_history(request):
     def side_effect(name):
-        print("history('{}')".format(name))
-        count = 2 if name == 'not-really-used' else 12
+        return 2 if name == 'not-really-used' else 12
     p = patch('thefuck.rules.no_command._count_history_uses',
               side_effect = side_effect)
     p.start()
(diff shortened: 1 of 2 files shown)


nvbn:
Thanks!

nvbn:
I checked changes locally and it's not working.
And maybe `history` is a bash/zsh command and can't be called simple from Popen.
Is it works for you?

jjmason:
It works locally for me, I'm on Ubuntu and zsh.  It's probably the history
thing, what environment have you been testing it under?

nvbn:
It's strange, I'm on ubuntu 14.10 and zsh too.
When I run `fuck` I've got:
```
/bin/sh: 1: history: not found
```

jjmason:
Oh I see what's going on, on my machine history is an alias to 'fc -l 1'.
I'll change it to use that, fc is a builtin so it should be available more
or less anywhere, although we should have a look at how to do it with fish
I suppose.

jjmason:
Fix pushed, should be on the PR unless you already merged it.

PR #43, opened 2015-04-19
Create rm_root.py

SpyCheese:
I think that it is a very useful rule.

spycheese@scc:~$ rm -r /
rm: it is dangerous to operate recursively on ‘/’
rm: use --no-preserve-root to override this failsafe
spycheese@scc:~$ fuck
rm -r / --no-preserve-root
rm: descend into write-protected directory ‘/’?

### DIFF
--- a/thefuck/rules/rm_root.py
+++ b/thefuck/rules/rm_root.py
@@ -1,6 +1,6 @@
 def match(command, settings):
-    return ('rm' in command.script
-            and '--help' not in command.script
+    return ('/' in command.script
+            and '--no-preserve-root' not in command.script
             and '--no-preserve-root' in command.stderr)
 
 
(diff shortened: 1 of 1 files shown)


JensTimmerman:
I would prefix with `sudo` to be sure?

nvbn:
I think this `rule` is too dangerous to be bundled with app =)

nwinkler:
:+1: If this ever gets bundled with the tool, please make it optional or disabled by default :-)

nvbn:
Added as disabled by default rule =)

CGreen313:
Thanks now I need to know how to use it the right way so I don't create more issue

CGreen313:
What ifs the best root to use with this in my galaxy s3

PR #46, opened 2015-04-19
sudo rule: add support for "Operation not permitted" errors

felixonmars:
`Operation not permitted` is also a commonly seen error that could be fixed by sudo in many cases. A few examples that this PR could fix:

```
$ hping -S -p 80 127.0.0.1
[open_sockraw] socket(): Operation not permitted
[main] can't open raw socket
```

```
$ sudo touch /tmp/tmp
$ chown nobody /tmp/tmp
chown: changing ownership of ‘/tmp/tmp’: Operation not permitted
```

```
$ umount /tmp
umount: /tmp: umount failed: Operation not permitted
```


nvbn:
Thanks!

PR #52, opened 2015-04-20
Added rule:cp_omitting_directory

NabeelValapra:
An Example:

➜ cp foo1/me/ foo2/
cp: omitting directory `foo1/me/'

➜ fuck
(would copy the directory by replacing 'cp' with 'cp -r' )

### DIFF
--- a/thefuck/rules/cp_omitting_directory.py
+++ b/thefuck/rules/cp_omitting_directory.py
@@ -1,0 +1,10 @@
+def match(command, settings):
+    if 'cp: omitting directory' in command.stderr.lower():
+        return True
+    return False
+
+
+def get_new_command(command, settings):
+    return command.script.replace('cp', 'cp -r') 
+
+


thefuck/rules/cp_omitting_directory.py
```diff
@@ -0,0 +1,10 @@
+def match(command, settings):
+    if 'cp: omitting directory' in command.stderr.lower():
+        return True
+    return False
+
+
+def get_new_command(command, settings):
+    return command.script.replace('cp', 'cp -r')
```
nwinkler: This is dangerous in the case the directory name contains the letters `cp`, they will be replaced as well. Either add the `1` parameter as a third parameter (`maxreplace`), or use a regular expression, e.g.

    return re.sub('^cp (.*), 'cp -r \\1')

nwinkler:
It would be great to have a test case as well...

nvbn:
Thanks!

NabeelValapra:
Thanks!, I'm on it.

PR #53, opened 2015-04-20
Added rm -rf support

nwinkler:
When someone tries to remove a directory without specifying `-rf`, e.g.

    $ mkdir foo
    $ rm foo

the shell refuses to remove the directory with an error:

    rm: foo: is a directory

Instead, you need to use

    $ rm -rf foo

This rule adds fuck support for this:

    $ mkdir foo
    $ rm foo
    rm: foo: is a directory
    $ fuck
    rm -rf foo


nvbn:
Thanks!

PR #54, opened 2015-04-20
Update README.md

olysyuk:
OSX has a different bash file
http://stackoverflow.com/questions/8967843/how-do-i-create-a-bash-alias


pottiep:
Ok thanks. Will try when im more sober.

Sent from Yahoo Mail on Android

nvbn:
Thanks!

PR #55, opened 2015-04-20
add a rule to run the script you enter by add './' in command's prefix

dannion:
add a rule to run your command by add './' in command's prefix if the command is a file exists in your path.

### DIFF
--- a/thefuck/rules/has_exists_script.py
+++ b/thefuck/rules/has_exists_script.py
@@ -1,0 +1,9 @@
+import os
+
+def match(command, settings):
+	exist = os.path.exists(command.script)
+	return exist
+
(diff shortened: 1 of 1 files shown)


thefuck/rules/has_exists_script.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+def match(command, settings):
+	exist = os.path.exists(command.script)
+	return exist
```
nwinkler: This could be merged into one line...

nwinkler:
:+1: Great idea!

A test case would be nice...

nvbn:
Thanks, I've added tests and fixed code style a bit.

PR #62, opened 2015-04-20
A special case for 'Permission denied' error msg when trying to execute ...

sjuvekar:
...a

python scripy.

The script does not have execute permission and/or does not start with !#/usr/...
In that case, pre-pend the command with 'python' keyword.

Change-Id: Idf73ee9cf0a523f51c78672188a457b2fcedc1e6

### DIFF
--- a/tests/rules/test_python_command.py
+++ b/tests/rules/test_python_command.py
@@ -1,0 +1,9 @@
+from thefuck.main import Command
+from thefuck.rules.sudo import match, get_new_command
+
+def test_match():
+    assert match(Command('', '', 'Permission denied'), None)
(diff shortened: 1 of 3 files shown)


tests/rules/test_python_command.py
```diff
@@ -0,0 +1,9 @@
+from thefuck.main import Command
+from thefuck.rules.sudo import match, get_new_command
```
nwinkler: Looks like the wrong import... Shouldn't you import `thefuck.rules.python_command` here?

sjuvekar: My bad, changed the commit

nvbn:
Thanks!

PR #64, opened 2015-04-20
mkdir -p

nwinkler:
When adding directories using `mkdir`, intermediate directories have to
exist, unless you specify the `-p` option:

    $ mkdir foo/bar/baz
    mkdir: foo/bar: No such file or directory
    $ fuck
    mkdir -p foo/bar/baz


nvbn:
Thanks!

PR #73, opened 2015-04-21
Added a string which could be thrown by Fedora's new dnf package manager.

Dugucloud:
"This command has to be run under the root user."


nvbn:
Thanks!

PR #78, opened 2015-04-21
adding a rule to run "sudo apt-get install"

luv:
A straight-forward rule; 

Catching all exceptions not to cause problems on systems without apt-get (or without CommandNotFound package)

### DIFF
--- a/thefuck/rules/apt_get.py
+++ b/thefuck/rules/apt_get.py
@@ -1,0 +1,28 @@
+import sys
+
+def match(command, settings):
+    try:
+        import CommandNotFound
+        if 'not found' in command.stderr:
+            try:
+                c = CommandNotFound.CommandNotFound()
+                pkgs = c.getPackages(command.script.split(" ")[0])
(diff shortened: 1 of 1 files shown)


nvbn:
Doesn't work on ubuntu 14.10.

From which package I can get `import CommandNotFound`?

luv:
Strange. On Ubuntu it should work "out of the box".

The package is "python-commandnotfound"

nicholerichardson:
I don't know exactly who is using my account this is Nicole Richardson but I am catching on and I do have your information downloaded it will not be long and I will know so stop !!! can anyone help ;???

nicholerichardson:
there is some video information I need to download that I have picked up from another phone I cannot open the application I need to download something from Apple for Android incredible two tell me what I need to download this information .

nvbn:
I've changed it a bit, so now it requires updated alias: `alias fuck='eval $(thefuck $(fc -ln -1))'`.

PR #82, opened 2015-04-21
add thing for when known hosts have changed

nicwest:
so I get messages like this all the time when I'm testing deployments:

```
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
Someone could be eavesdropping on you right now (man-in-the-middle attack)!
It is also possible that a host key has just been changed.
The fingerprint for the RSA key sent by the remote host is
b6:cb:07:34:c0:a0:94:d3:0d:69:83:31:f4:c5:20:9b.
Please contact your system administrator.
Add correct host key in /Users/nic/.ssh/known_hosts to get rid of this message.
Offending RSA key in /Users/nic/.ssh/known_hosts:74
RSA host key for project.local has changed and you have requested strict checking.
Host key verification failed.
```
I don't want to turn off strict checking, but it's also a minor pain to go to known hosts and delete the conflicting thing in person.

I'm also thinking that the obvious partner to this rule would be one that chmods keys when you forget.

### DIFF
--- a/tests/rules/test_ssh_known_host.py
+++ b/tests/rules/test_ssh_known_host.py
@@ -1,0 +1,74 @@
+import os
+import pytest
+from thefuck.main import Command
+from thefuck.rules.ssh_known_hosts import match, get_new_command, remove_offending_keys
+
+
+@pytest.fixture
+def ssh_error(tmpdir):
+    path = os.path.join(str(tmpdir), 'known_hosts')
+
+    def reset(path):
+        with open(path, 'w') as fh:
+            lines = [
+                '123.234.567.890 asdjkasjdakjsd\n'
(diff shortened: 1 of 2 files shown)


nvbn:
Thanks!

I think for permissions to `.ssh` there should be separate rule.

PR #93, opened 2015-04-22
Used re.sub() instead of .replace()

NabeelValapra:


nvbn:
We don't need to use `-r` option, `-a` option includes it.

PR #95, opened 2015-04-22
Added colorama in requirements.txt

Dugucloud:


nvbn:
It shouldn't be in `requirements.txt`, it's already in [setup.py install_requires](https://github.com/nvbn/thefuck/blob/master/setup.py#L18).

Are `setup.py develop/install` installs it for you?

PR #96, opened 2015-04-22
Making a confirm before executing fixed command

599316527:
My colleague is concerned that the fixed command is not the one he want to execute and even damage files. A confirm before really executing is necessary. @Justineo


nvbn:
Try latest master and alias:
```bash
alias shit='$(THEFUCK_REQUIRE_CONFIRMATION=true thefuck $(fc -ln -1))
```
I think it does what you want.

599316527:
Cool! It works well.

Justineo:
Great.

PR #99, opened 2015-04-22
fix rm dir rule to make it case insensitive

timofurrer:
In bash the output for the command `rm -f foo/` is:

    rm: cannot remove ‘foo/’: Is a directory

And not:

    rm: cannot remove ‘foo/’: is a directory

### DIFF
--- a/tests/rules/test_rm_dir.py
+++ b/tests/rules/test_rm_dir.py
@@ -4,6 +4,7 @@ from thefuck.rules.rm_dir import match, get_new_command
 
 def test_match():
     assert match(Command('rm foo', '', 'rm: foo: is a directory'), None)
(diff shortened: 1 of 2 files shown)


nwinkler:
Good catch! Seems to be OS-specific. On OS X (10.10.3, GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)), it's using the lower case version (_is a directory_).

timofurrer:
Seems like. I have:

            Linux XXX 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt2-1 (2014-12-08) x86_64 GNU/Linux

nvbn:
Thanks!

PR #106, opened 2015-04-23
New rule: sl -> ls

Brobin:
When typing quickly, I make this mistake all the time. This addition allows you to unfuck sl, replacing it with ls.


nwinkler:
Typing `ls` again is two characters, typing `fuck` is four characters that you have to type. Isn't it quicker to simply type `ls`?

Brobin:
Yeah, but that's not as satisfying! The `cd..` - > `cd ..` rule doesn't save characters either.

nwinkler:
True...

Dugucloud:
This mistake is truly fuckable!!!

SanketDG:
I somehow believe that the alias created for `thefuck` should be way small, a maximum of three characters. `fuck` just seems way too big.

nvbn:
Thanks!

PR #107, opened 2015-04-23
Add a support for pip unknown commands

kimtree:
This is a initial support for PIP (the most popular package install in python as you already know).


Dugucloud:
Cool, I think apt, yum and dnf will like it.

nvbn:
Thanks!

PR #110, opened 2015-04-23
Support brew unknown command

kimtree:
This also same as pip unknown command. ( issue #107 )

This patch has a command dictionary for HomeBrew 0.9.5 to recommend commands for unknown commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ sudo pip install thefuck --upgrade
 The Fuck tries to match rule for the previous command, create new command
 using matched rule and run it. Rules enabled by default:
 
+* `brew_unknown_command` &ndash; fixes wrong brew commands, for example `brew docto/brew doctor`;
 * `cd_parent` &ndash; changes `cd..` to `cd ..`;
 * `cp_omitting_directory` &ndash; adds `-a` when you `cp` directory;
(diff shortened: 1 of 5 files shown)


nwinkler:
Instead of providing a dictionary, wouldn't it make more sense to use something like the existing Bash completion for Homebrew?

    `brew --prefix`/Library/Contributions/brew_bash_completion.sh

This file has a complete definition of all of the available brew commands.

kimtree:
@nwinkler AFAIK, That script only supports a command which starts with original command (e.g. inst -> install). But the most case users use 'the fuck' command, they expect this to support auto-correct as well. In the code I wrote, It will automatically finds the most similiar command which user types.

Also I know that using static dictionary in the code is bad for working with several user environment. It would be great if somebody suggest other method to support auto-correct as well.

nwinkler:
@kimtree Sure, I understand that. I was trying to suggest that instead of hardcoding the possible commands in the rule, maybe there's a way to use the existing completion command to get a hold of the possible commands supported by the locally installed version of Homebrew. I don't know how to go about this, but wanted to make the suggestion nonetheless.

kimtree:
@nwinkler How about see the file name lists in https://github.com/Homebrew/homebrew/tree/master/Library/Homebrew/cmd ?

nvbn:
Thanks!

PR #112, opened 2015-04-23
Using eval for Bash alias

### LINKED ISSUE #108: Doesn't handle pipes
It looks like `fuck` doesn't handle pipes well. If I try to do this in Bash:

```bash
$ pss -ef | grep foo
-bash: pss: command not found
$ fuck
ps -ef | grep foo
ps: illegal argument: |
usage: ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]]
          [-g grp[,grp...]] [-u [uid,uid...]]
          [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
       ps [-L]
```

It looks like the command is executed in the wrong way, it seems like it's trying to execute it as a single command, instead of treating it as a line of input with potentially multiple commands.

nwinkler:
This fixes #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ brew install thefuck
 And add to `.bashrc` or `.zshrc` or `.bash_profile`(for OSX):
 
 ```bash
-alias fuck='$(thefuck $(fc -ln -1))'
-# You can use whatever you want as an alias, like for mondays:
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

PR #114, opened 2015-04-23
Added rule for fixing Alt+Space character

nwinkler:
Happens on the Mac a lot when typing a pipe character (Alt+7), and keeping the Alt key pressed down for a bit too long, so instead of Space, you're typing Alt+Space. This rule replaces the Alt+Space with a
simple Space character.

```bash
$ ps -ef | grep foo
-bash:  grep: command not found
$ fuck
ps -ef | grep foo
    502 29003  6587   0  3:18pm ttys007    0:00.00 grep --color=auto foo
```


nvbn:
Thanks!

PR #118, opened 2015-04-24
Adds a fuck for ls that adds -lah

crimsonknave:
This could easily be configurable, but I didn't see any similar settings (but I could easily have missed them).


nvbn:
Thanks!
Added as disabled by default, because it's a bit not how the `fuck` should work =)

PR #119, opened 2015-04-24
conf: initialize a settings file if it doesn't exist (fix #111)

### LINKED ISSUE #111: settings.py is empty after installation
Please add content to settings.py for those who do not know python, so they can edit it.

scorphus:
Please review and comment. Thanks!


nvbn:
Thanks!

Issue #98, opened 2015-04-22
What command line interpreter used for rules


ncrocfer:
Hello,

I want to create a new rule based on the `cd` command, but I noticed that the returned error is not the same when I use bash or sh.

With bash :
```
$ cd foo
bash: cd: foo: No such file or directory
```
With sh :
```
$ cd foo
sh: 1: cd: can't cd to foo
```

And when I print `command.stderr` in the `match()` function, the stderr is the sh style. So my question is : when we create a new rule, should we use this stderr style to test errors?

Thank you :smiley:

nwinkler:
Looks like #50...

nvbn:
Python uses `/bin/sh` as a shell for `subprocess.Popen` and related functions.
We should use output from `/bin/sh`.

ncrocfer:
Ok, thanks !
My rule is the same as #50, so I'm waiting his PR rather than sending mine.

PR #120, opened 2015-04-24
Added cd_mkdir rule

### LINKED ISSUE #50: cd to mkdir
I'm trying to create a fuck for `cd: no such file or directory: <directory_name>` which gets converted to mkdir <directory_name>. 
I wrote following for match function but is says `No fuck given`
```
def match(command, settings):
    return ("cd: no such file or directory:" in command.stderr)
```
I tried printing stderr which contains `/bin/sh: 1: cd: can't cd to <directory_name>`. Why is it so and where do I get "cd: no such file or directory:"

nwinkler:
This fixes #50 and #98.

```bash
$ cd foo/bar/baz
cd: foo: No such file or directory
$ fuck
mkdir -p foo/bar/baz && cd foo/bar/baz
```

Added matchers for both Bash and sh error messages. Depending on your
default shell, the messages might be slightly different.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ using matched rule and run it. Rules enabled by default:
 
 * `brew_unknown_command` &ndash; fixes wrong brew commands, for example `brew docto/brew doctor`;
 * `cd_parent` &ndash; changes `cd..` to `cd ..`;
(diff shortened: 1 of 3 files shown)


klinkin:
+1

nvbn:
Thanks!

Issue #131, opened 2015-04-26
Add a command to output the bash alias


WhoBrokeTheBuild:
The idea would be to do something like

thefuck --bash-alias >> ~/.bashrc

nvbn:
`thefuck-alias` added.

PR #132, opened 2015-04-26
Added the --bash-alias command to output the alias

WhoBrokeTheBuild:
Added the following command output the correct alias line to add to your .bashrc so you can call it like:
thefuck --bash-alias >> ~/.bashrc

This would be the implementation for issue #131

### DIFF
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -111,11 +111,20 @@ def is_second_run(command):
     """Is it the second run of `fuck`?"""
     return command.script.startswith('fuck')
 
+def check_show_alias(argv):
+    """Was thefuck called with --bash-alias?"""
+    if "--bash-alias" in sys.argv:
+        print("\nalias fuck='eval $(thefuck $(fc -ln -1))'\n");
+        return True
+    return False
 
 def main():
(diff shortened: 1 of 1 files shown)


nvbn:
We can't use cli arguments in this app, reason why:
```bash
➜ someapp --bash-alias  
zsh: command not found: someapp
➜  fuck  # And it just sets new alias
➜ alias fuck
fuck=\''eval'
```

SanketDG:
[Docopt](https://github.com/docopt/docopt) maybe? I don't know if it would work out, as @nvbn pointed out.

nvbn:
@SanketDG docopt, optparse and argparse can't help.

@WhoBrokeTheBuild  only way to do this properly - introduce new `entry_point`, like `thefuck-alias`.

SanketDG:
@nvbn That would be silly though, anybody would hardly use it that way.

nvbn:
@SanketDG what difference between `thefuck --alias` and `thefuck-alias`?

SanketDG:
Well people are generally habituated to use command line arguements, but I guess this would be a workaround.

PR #133, opened 2015-04-26
Add tip on installing requirements on Ubuntu

firstdoit:
In trying to avoid that people leave the guide searching for how to install the requirements, I think it's useful to provide a simple command.

Perhaps we could add also commands for OS X and Fedora?

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,12 @@ Reading package lists... Done
 - python
 - python-dev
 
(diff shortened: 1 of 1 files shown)


nvbn:
I moved os specific installation steps to the [wiki](https://github.com/nvbn/thefuck/wiki/Installation), it'll be good to add commands for other OS there (OS already added).

PR #139, opened 2015-04-28
Add a support for git command alish as well

kimtree:
This patch quite tricky to support alias command. Please close If you think this is not good.


nvbn:
Hardcoded aliases isn't a solution, I'll try to add support of aliases in nearest time.

kimtree:
@nvbn :thumbsup:

PR #145, opened 2015-04-28
Adds example lines of settings.py in readme.py

daDnaK:
To mitigate an empty settings file and as in the readme was no specification of the format, these are some sample lines on how to format the settings file


nvbn:
It's not how it works =)
But thanks, I added proper examples in the readme.

daDnaK:
well at least now I know how it's done, thanks! ;-)

PR #146, opened 2015-04-29
Improve a logic to get recommended command based on local environment

kimtree:
I made this function to recommend command based on static commands for the first time. But now I changed it to look into local folders to get available commands for brew and brew taps.


nvbn:
Thanks!

PR #149, opened 2015-04-30
Add sudo rule for shutdown command.

SanketDG:
image

This rule adds the terminal output on running `shutdown` without `sudo`.


nvbn:
Thanks!

PR #158, opened 2015-05-03
Update sudo.py

kirillDanshin:
Add a new pattern for dpkg on Russian version of Ubuntu/Debian.
What you think about that?


nvbn:
There's no need in localized patterns, `thefuck` works with commands output with `LANG=C`.

PR #159, opened 2015-05-05
Add a rule for pacman

mcarton:
This adds a rule for `pacman`, Arch Linux's package manager.

It is similar to the *apt-get* rule. It is enabled by default only if `pkgfile` is available and prefers `yaourt` (a wrapper for `pacman`) over `pacman` itself if available.


nvbn:
Thanks!

PR #160, opened 2015-05-05
Follow the XDG Base Directory Specification

mcarton:
Namely, put configuration files in `~/.config`.

It should not break already existing configuration as `~/.thefuck` is used if it exists.


mcarton:
I just fixed a conflict. Any reason not to merge this one?

PR #162, opened 2015-05-05
docker thefuck

brunoborges:


nvbn:
It's already handled by `permission denied` pattern in `sudo` rule.

brunoborges:
Oh, cool thanks.

PR #164, opened 2015-05-06
Gave all modules priority over no_command

atrakh:
This fixes a bug with the no_command module being used instead of sl_ls for the input of 'sl'.


nvbn:
Implemented by internal priority mechanism in 1.36.

PR #165, opened 2015-05-06
rules(pacman): make it compatible with Python 2

scorphus:
Tests are on the way. As soon as I can.

### DIFF
--- a/thefuck/rules/pacman.py
+++ b/thefuck/rules/pacman.py
@@ -1,8 +1,17 @@
+import six
 import subprocess
+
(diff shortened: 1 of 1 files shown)


nvbn:
Maybe it's better to not monkeypatch subprocess module and just use:
```python
DEVNULL = open(os.devnull, 'w')
```
It works on both python versions.
Also, `FileNotFoundError` is inherited from `OSError`, so maybe just handling `OSError` will be ok.

nvbn:
Fixed =)

scorphus:
Right on!

Issue #163, opened 2015-05-05
Idea: let thefuck learn from your fuckups


underscorejho:
It would be nice if, instead of hard-coding possible wrong commands so much, there was some sort of `add_fuckup` command. Then in the future, there would be a rule or something that acted as a rule that watched for that fuckup.

For instance:

Instead of hard-coding the `cd..` fuckup, when you fuck up for the first time...

    -$ cd..
    cd..: command not found

and get...

    -$ fuck
    no fucks given

you have the option to...

    -$ fuck --add_fuckup
    thefuck added fuckup cd.. on cd

... or something like that.

nwinkler:
How would it know what to do? How would it know to run `cd ..` instead of `cd..` in this case?

underscorejho:
A dictionary maybe? Where the key is the fuckup and the value is the correct command. Probably saved locally somewhere. 

I'm really more interested in seeing if this is an idea people would be interested in. If so, then I can certainly do more research into the most effective ways to implement it. It just seems like a lot of hassle to have to manually enter rules for possible errors.

uGuardian:
Couldn't you make it also that if you use fuck and it doesn't come up with something, you could reenter the command correctly and then use the add fuckup command to add the most recent command as a correction so you wouldn't have to manually hard-code the explicit correction?

nwinkler:
That sounds like a reasonable approach! It would need some fiddling with the command parameters so e.g. learning that the `foo/bar/baz` part in `mkdir foo/bar/baz` is the dynamic part, so when you run `mkdir this/is/great` and you get the similar error message to the first command, it will know what to do.

Many of the rules are built in the following fashion:

* Apply regex match on `command.script`, `command.stdout` and `command.stderr`.
* Replace/transform part of `command.script` to fix the issue, e.g. prepending `sudo` or adding a parameter (`-p` in the `mkdir` case).

I agree that there should be a moderately easy way to define these without programming the rules...

underscorejho:
Sweet. 

Ok. So on this feature-idea here we would want:

* local fuckup learning using an `add_fuckup` command (or `addfuckup`.. or `newfu`... or something)
* confirmation that it's learning the right fuckup/command pair (but this could be optional)
* probably some `learned_fuckups` rule that handles these new learned fuckups
* ... other things?

nwinkler:
A way to share learned fuckups?

The learned fuckups should be stored in the local `~/.thefuck` folder.

[...]

PR #167, opened 2015-05-07
custom fuckups with --ifuckedup

underscorejho:
Added learning ability to `thefuck`. #163 

```
-$ cd..
cd..: command not found
-$ fuck --ifuckedup 'cd ..'
Adding fucked: cd.. as fixed: cd ..
-$ fuck --remove 'cd..'
Removing fucked: cd.. as fixed: cd ..
```


underscorejho:
fixing bugs...

sullyj3:
Personal opinion: I prefer your original name for the option, "add_fuckup", over "ifuckedup". "add_fuckup" makes it clearer what it's for. Whereas "ifuckedup" is ambiguous, you fucked up every time you need to type fuck, not just when you need to add a new rule.

underscorejho:
You make a good point. If @nvbn agrees, I'll change that.

nvbn:
It's a cool idea, but we can't use command line arguments, because previous command can be `cmd --remove` or something, and for this situation `fuck` wouldn't work.

underscorejho:
I'm going to make some changes first to make sure that this is true, but I believe I can build it so that it's only looking at command line args from the relevant `fuck` command, instead of the 'fucked up' command. That way `fuck --remove` wouldn't interfere with `cmd --remove`.

underscorejho:
I've changed the get_command() function in main.py to programmatically fetch the previous command using .bash_history . This eliminates the need for the current alias, though it requires lines in .bashrc to keep .bash_history up to date. 

I believe this should work, but I'm having trouble adjusting the appropriate tests.

sullyj3:
Does it work with zsh and fish?

PR #171, opened 2015-05-07
Add a don't repeat yourself rule

mcarton:
How often did I start typing a command, though about what I actually wanted to do and wrote that command from the start, ending with commands like `git git commit …`?

No more! This rule removes the first word in case of such repetitions.


nvbn:
Thanks!

PR #183, opened 2015-05-12
fix(brew_unknown_command): make subprocess.check_output return str

scorphus:
Fix `TypeError: can't concat bytes to str` error on Python 3.4.


nvbn:
Thanks!

Issue #184, opened 2015-05-12
thefuck-alias writes to stderr


and-dmitry:
thefuck-alias seems to write to stderr instead of stdout, so the command from README does not change `.bashrc`:

```
$ thefuck-alias >> ~/.bashrc

alias fuck='eval $(thefuck $(fc -ln -1))'
```

Of course one could just type `$ thefuck-alias 2>> ~/.bashrc` but it doesn't feel right.

SanketDG:
I implemented `thefuck-alias` and I am pretty sure it works. It simply prints out in the terminal, so it should go to to `stdout`.

SanketDG:
Thanks for pointing that out! I had implemented `thefuck-alias` and I must have missed it through.

SanketDG:
Um, yeah, it is because it was being returned through a function. Will work okay after #185 is merged.

nvbn:
Currently it writes to stdout:
```bash
➜  thefuck-alias > out
➜  cat out

TF_ALIAS=fuck alias fuck='eval $(thefuck $(fc -ln -1 | tail -n 1)); fc -R'

```

PR #185, opened 2015-05-12
Fix thefuck-alias to print in app_alias function.

SanketDG:
The issue was returning and not printing. Reported #184


nvbn:
Thanks!

SanketDG:
Yeah, I guess returning from a function and then printing is much better. Sometimes I just forget about Python 3. Need to get that into my workflow.

PR #187, opened 2015-05-12
Add a git_stash rule

mcarton:


nvbn:
Thanks!

thefuck/rules/git_stash.py
```diff
@@ -0,0 +1,7 @@
+def match(command, settings):
+    return ('git' in command.script
+            and 'Please commit or stash them.' in command.stderr)
+
+
+def get_new_command(command, settings):
+    return 'git stash && ' + command.script
```
nwinkler: Shouldn't it unstash the changes after running the command? Like `git stash pop`?

thefuck/rules/git_stash.py
```diff
@@ -0,0 +1,7 @@
+def match(command, settings):
+    return ('git' in command.script
+            and 'Please commit or stash them.' in command.stderr)
```
dunn: Could this be tweaked to work for switching branches too?  That error message is `Please, commit your changes or stash them before you can switch branches.` So maybe the `match` could be as simple as `or stash them`.

mcarton:
@nicwest: well, I don't like `git stash pop`; see https://codingkilledthecat.wordpress.com/2012/04/27/git-stash-pop-considered-harmful/ and plenty others.
@dunn: sure, I will do that tomorrow.

nicwest:
@nwinkler: I think he meant that for you.

PR #190, opened 2015-05-13
Add TCSH support

HughMacdonald:
Add tcsh support.

Currently doesn't properly support history re-writing, though.


nvbn:
Thanks!

PR #192, opened 2015-05-13
Fix brew_install rule on Python 3

scorphus:
With this change, the `brew_install` rule works on Python 3. Also, there are no more tests being ignored – there were two of them.


nvbn:
Thanks!

PR #193, opened 2015-05-13
Custom fuckups rule -- no command line args

underscorejho:
version of #167 without command line args


nvbn:
Why you use uppercase for variables?

underscorejho:
I guess that's just my style. Upper case variables are more or less 'constants'. I can change that if you like.

nvbn:
Nah, I'll change it by myself. I'll add support of `prepare` functions to rules and merge this pull request in nearest time =)

Also, there's convention in python community that variable written in uppercase is a constant, but it's not like `const` or `let`, it's more like `define`. So most of time it's a bad practice to use uppercase variable =)

underscorejho:
Sweet. Thanks for the tip.

PR #197, opened 2015-05-15
Add a `whois` rule

mcarton:
What the `whois` command returns depends on the 'Whois server' it contacted and is not consistent through different servers. But there can be only two types of errors I can think of with `whois`:
- `whois https://en.wikipedia.org/` → `whois en.wikipedia.org`;
- `whois en.wikipedia.org` → `whois wikipedia.org`.

So we match any `whois` command and then:
- if there is a slash: keep only the FQDN;
- if there is no slash but there is a point: removes the left-most subdomain.

We cannot either remove all subdomains because we cannot know which part is the subdomains and which is the domain, consider:
- www.google.fr → subdomain: www, domain: 'google.fr';
- google.co.uk → subdomain: None, domain; 'google.co.uk'.

### DIFF
--- a/thefuck/rules/whois.py
+++ b/thefuck/rules/whois.py
@@ -1,0 +1,30 @@
+from urllib.parse import urlparse
+
+
+def match(command, settings):
+    """
+    What the `whois` command returns depends on the 'Whois server' it contacted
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

PR #203, opened 2015-05-16
ignore tag files

aragacalledpat:


mcarton:
How is that specific to thefuck? You should probably put that in your global *gitignore*.

aragacalledpat:
Sorry! I'm new to some of this stuff.  Didn't know I could do that.

PR #201, opened 2015-05-16
Add test and complete README for the `whois` rule

mcarton:

PR #204, opened 2015-05-16
Fixed a problem at whois.py

whopper666:
While the normal version of thefuck worked perfectly for me, the developer version did not. I had to add encoding line at rules/whois.py and also change the source of urlparse module. As a newbie i don't expect this solution to be perfect, it just worked for me and I wanted to bring this problem to your attention.

PS It's my first pull request ever so I would appreciate the feedback if something was terribly wrong

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 
-VERSION = '1.40'
+VERSION = '1.41'
 
 
 setup(name='thefuck',
--- a/thefuck/rules/whois.py
+++ b/thefuck/rules/whois.py
@@ -1,4 +1,5 @@
-from urllib.parse import urlparse
+ # coding: utf-8
+from urlparse import urlparse
 
 
 def match(command, settings):


scorphus:
I appreciate your intent to contribute! But this issue was already reported and a solution already suggested, see #201.

Also see my line notes.

mcarton:
Those are 2 problems with Python 2. It works perfectly on Python 3 and I admit I did not test it with Python 2 when I made the first pull request (see below). Note that commit 2705ff4 no breaks Python 3 and this pull request should not be merged. f04c439 from #201 fixes both problems.

I though people still used Python 2 because of big and complicated libraries without Python 3 port or for big projects that would be long to port. This is not the case for `thefuck`, so why use Python 2? (This is a real question, Python is not my preferred language and I actually wonder why this project bother with Python 2 compatibility)

setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 
-VERSION = '1.40'
+VERSION = '1.41'
```
scorphus: Bumping new releases is up to project owners/maintainers. Also, it doesn't really belongs to any specific fix/feature one tries to solve/implement.

I believe you changed it in order to have it upgraded by `pip install`. I'm not sure about your experience with Python development, but if you allow me to give you what may be a tip, you can use `pip install -U -e .` from the root of the project to have it installed in *editable mode*:

```
$ pip help install
[snip]
Install Options:
  -e, --editable <path/url>   Install a project in editable mode (i.e. setuptools "develop mode")
                              from a local project path or a VCS url.
[snip]
  -U, --upgrade               Upgrade all specified packages to the newest available version. This
                              process is recursive regardless of whether a dependency is already
                              satisfied.
[snip]
```

Also, allow me to strongly recommend you using [Virtualenv](http://virtualenv.readthedocs.org/en/latest/), if you're not already doing so

thefuck/rules/whois.py
```diff
@@ -1,4 +1,5 @@
-from urllib.parse import urlparse
+ # coding: utf-8
```
scorphus: The actual line should be:
```python
# -*- encoding: utf-8 -*-
```

thefuck/rules/whois.py
```diff
@@ -1,4 +1,5 @@
...
+from urlparse import urlparse
```
scorphus: You made this line work with Python 2, but now it fails with Python 3. The best option is to use [six](https://six.readthedocs.org/). Check #201.

PR #205, opened 2015-05-17
Adding cd correction rule to the default rules

mmussomele:
Hi there! I just want to start off by saying that thefuck is great and it has saved me a lot of time. That being said, one thing that has bothered me is that whenever I mess up a cd, the cd_mkdir rule just creates a directory that 99% of the time was just a typo and then I have to step out of it and remove it. 
So I went ahead and wrote this rule, which goes through a cd command and attempts to spellcheck it (relative to the user's actual folders), fixing any suspected typos along the way. I hope you like it!


underscorejho:
This is awesome. But just a note, the way I usually avoid these typos in my day to day use is to use tab to autocomplete. Then, if it can't autocomplete, I've clearly mistyped it. 

But in terms of this project, this is really good way to handle this.

nwinkler:
This is great! I wonder though whether you have to provide your own implementation of finding close matches.

The `no_command` rule uses a library function to do this, maybe you could reuse that: https://github.com/nvbn/thefuck/blob/master/thefuck/rules/no_command.py#L24

mmussomele:
Good catch! I'll swap that in instead!

nvbn:
Thanks!

PR #212, opened 2015-05-20
fix(whois): check if there's at least one argument to `whois`

scorphus:
This avoids thefuck failing when there's no arguments. It fails with:

```
  ...
  File "thefuck/rules/whois.py", line 26, in get_new_command
    url = command.script.split()[1]
IndexError: list index out of range
```


nvbn:
Thanks!

PR #214, opened 2015-05-21
refact(man): do not match if there's no argument to man

scorphus:
If there's no argument to man, a call to thefuck should just give no fuck, like the following:
```bash
~> man
What manual page do you want?
~> fuck
No fuck given
```


nvbn:
Thanks!

Issue #210, opened 2015-05-20
correct 'open' command on Mac to open browser


zag2art:
```bash
zag2art@laptop:~/workfiles/docclean$ open github.com
The file /Users/zag2art/workfiles/docclean/github.com does not exist.
Perhaps you meant 'http://github.com'?
zag2art@laptop:~/workfiles/docclean$ fuck
No fuck given
```
Can someone add the case into rules? It should fire `open http://github.com` in my case. Sorry, don't know python.

scorphus:
This issue is fixed by https://github.com/nvbn/thefuck/commit/d3146aa0aca57522820f5bdb0a22f5b248de407e and can be closed.

PR #216, opened 2015-05-23
Addressed Issue #210

diezcami:
Hi, I wanted to address the issue regarding the `open` command for URL's not being recognised by fuck.
The code I provided works for the more popular domain extensions and the tests. However, I actually originally wanted to do something along these lines:
``` python
def match(command, settings):
	#Unpopular Extensions
	ext = ['.ly', '.se', '.io', '.edu']
	return (command.script.startswith ('open')
			and 'does not exist' in command.stderr
			and ('http' in command.stderr
			or any(e in command.script for e in ext)))
```
I couldn't get the `any` syntax or anything that involves `command.stderr` to work with my code while testing though. It'd be great if you could shed some light on this issue for me so I could edit the code :) Thanks!

### DIFF
--- a/tests/rules/test_open.py
+++ b/tests/rules/test_open.py
@@ -1,0 +1,25 @@
+import pytest
+from thefuck.rules.open import match, get_new_command
+from tests.utils import Command
+
+
+@pytest.mark.parametrize('command', [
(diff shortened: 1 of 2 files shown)


nvbn:
Thanks!

PR #219, opened 2015-05-26
fix(shell::Fish): avoid looping when calling `fuck` twice

scorphus:
Or whatever the `thefuck` function name is.


scorphus:
Still got something more to do to avoid #221... working on it.

nvbn:
Thanks!

Issue #222, opened 2015-05-26
thefuck doesn't work with an alias that starts with zsh shell keyword


emptyaitch:
I have the following alias in my .zshrc:

alias mkdir='nocorrect mkdir -p'

`thefuck mkdir /nopermissions` uses the no_command rule, because running it through /bin/sh outputs "command not found" for nocorrect (a zsh keyword)

scorphus:
This is related to #262 and its fix should apply to this as well. I'll soon open a pull request.

PR #223, opened 2015-05-26
Fixing bug with /bin/sh not understanding other shells' keywords

emptyaitch:
This is meant to fix the following issue:

https://github.com/nvbn/thefuck/issues/222

It reads the os.environ["SHELL"] to run the command to get the output.  If SHELL isn't defined, os.environ.get() should return None, which will run the default "/bin/sh" .

### DIFF
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -79,7 +79,8 @@ def get_command(settings, args):
         return
 
     script = shells.from_shell(script)
-    result = Popen(script, shell=True, stdout=PIPE, stderr=PIPE,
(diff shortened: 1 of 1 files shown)


emptyaitch:
The corrects the interpretation, but still has bugs when using "sudo"

emptyaitch:
I solved my specific "nocorrect" problem by writing a new sudo.py rule that strips out the "nocorrect" and "noglob" keywords.   

nocorrect mkdir -p /nopermissions --> sudo mkdir -p /nopermissions

This is a dumb solution, because it is zsh-specific.

PR #224, opened 2015-05-27
Added javac rule

diezcami:
Fixes `javac` commands without `.java` file extensions
```
$ javac foo
error: Class names, 'foo', are only accepted if annotation processing is explicitly requested
```


nvbn:
Thanks!

PR #225, opened 2015-05-27
Added java rule

diezcami:
Removes `.java` extension when running Java programs
```
$ java foo.java
Error: Could not find or load main class foo.java
```


nvbn:
Thanks!

PR #227, opened 2015-05-28
Added a sudo string of Fedora's fedup

Dugucloud:
Only the last commit is effective. Other commits I created were to synchronize with nvbn's repo (sorry I'm not quite familiar with git)


nvbn:
Thanks!

PR #232, opened 2015-05-30
Wrap apt-get rule in sudo_support

bugaevc:
Fixes `sudo_support` not working for `no_command` rule.

Example:

```
$ sudo apt0get update
sudo: apt0get: command not found
$ fuck
```

would try to do `sudo apt-get install sudo && sudo apt0get update` and `no_command` rule wouldn't be executed at all.

By the way, there are some other rules that are not wrapped in `sudo_support` (ex. git rules). Would it make sense to wrap them too?


nvbn:
Thanks!

PR #236, opened 2015-06-01
Git branch list

bugaevc:
Just catches `git branch list` in place of `git branch` - useful for those who are confused by `git stash` vs `git stash list`.


nvbn:
Thanks!

PR #240, opened 2015-06-02
Added brew_upgrade rule

diezcami:
Appends `--all` to `brew upgrade`. Not relevant yet, but I've been getting errors that Homebrew will be disabling `brew upgrade` without flags eventually, so I thought I'd add it here for good measure.


nvbn:
Thanks!

PR #249, opened 2015-06-06
Add two cargo related rules

mcarton:
Adds two rules for `cargo`:

- `cargo`: for those who are too used to simply type `make` and are fucked every time they want `cargo` to actually compile their project;
- `cargo_no_command`: fixes spelling mistakes, similar to the `git_not_command` rule.


nvbn:
Thanks!

PR #252, opened 2015-06-11
Use an environment variable that keeps the alias name to avoid looping

scorphus:
@nvbn What do you think? Please review. Any comment is welcome :smiley:


nvbn:
Thanks!

PR #257, opened 2015-06-16
Fix sudo rule

tjhorner:
This fixes the sudo rule so when a command outputs something like `You need to be root to perform this command.` (like yum), it will detect it.

EDIT: doesn't work, hold on. adding more changes.

EDIT 2: committed the test, but not the rule lol


nvbn:
Thanks!

Also, sudo patterns are case insensitive.

tjhorner:
That's odd... when I tested the "need to be root" rule locally it didn't
pass. Must have been a typo when I tested it. (I didn't use the actual
test, I used `yum` without sudo)

PR #264, opened 2015-06-26
Cleanup

mcarton:
Some cleanup. 40fe604 fixes a problem with `fish`. b59e83c annoyed me for a while :) As for 7173e0d, I came across a tab in a file and just had to replace them all.

Note: This concerns many different things, tell me if you'd prefer to have different PR.


nvbn:
Wow! Thanks!

PR #265, opened 2015-06-26
Open

mcarton:
This is far from perfect but I can't think of a clean way to test if `foo.bar` can be a domain name, especially now that we allow things like `.paris` or `.diamonds`.
The complete list seems to [be available](https://www.iana.org/domains/root/db) so we could use that instead of the way too short list we currently have, but then we'd still have to differentiate the python file `foo.py` from a Paraguayan website.

### DIFF
--- a/thefuck/rules/open.py
+++ b/thefuck/rules/open.py
@@ -8,7 +8,7 @@
 
 
 def match(command, settings):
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

Issue #260, opened 2015-06-23
Fuck not working on linux


zakiaziz:
I've installed it into my mac and works just fine. But running it on a linux shell I've SSH'd into returns this:

[user@machine]$ git statis
git: 'statis' is not a git command. See 'git --help'.

Did you mean this?
	status
[user@machine]$ fuck
Traceback (most recent call last):
  File "/usr/bin/thefuck", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python2.6/site-packages/pkg_resources.py", line 2655, in <module>
    working_set.require(__requires__)
  File "/usr/lib/python2.6/site-packages/pkg_resources.py", line 648, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python2.6/site-packages/pkg_resources.py", line 546, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: colorama

SanketDG:
How did you install it?

zakiaziz:
`sudo pip install thefuck`

then `vi ~/.bashrc` and added this line: `alias fuck='eval $(thefuck $(fc -ln -1)); history -r'`

SanketDG:
What OS are you on?

SanketDG:
You can try `pip install --upgrade colorama` and then see if it works.

zakiaziz:
My OS info: centos-release-6-6.el6.centos.12.2.x86_64

`pip install --upgrade colorama` did not solve the issue for me

nvbn:
App requires python 2.7+ or 3.3+.
Python 2.6 isn't supported.

zakiaziz:
Thanks @nvbn

Issue #142, opened 2015-04-28
getting SyntaxError: invalid syntax


kj384g:
[kj384g@zion.np.wc1 ~]$ /usr/bin/thefuck
Traceback (most recent call last):
  File "/usr/bin/thefuck", line 9, in <module>
    load_entry_point('thefuck==1.30', 'console_scripts', 'thefuck')()
  File "/usr/lib/python2.6/site-packages/pkg_resources/__init__.py", line 546, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.6/site-packages/pkg_resources/__init__.py", line 2666, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.6/site-packages/pkg_resources/__init__.py", line 2339, in load
    return self.resolve()
  File "/usr/lib/python2.6/site-packages/pkg_resources/__init__.py", line 2345, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python2.6/site-packages/thefuck/main.py", line 9, in <module>
    from . import logs, conf, types
  File "/usr/lib/python2.6/site-packages/thefuck/conf.py", line 56
    for key in DEFAULT_SETTINGS.keys()
      ^
SyntaxError: invalid syntax

new to python, not sure what to check. Thanks!

nvbn:
Only python 2.7 and 3.3+ are supported, 2.6 - not.

PR #267, opened 2015-07-01
add python version testing in setup.py

SanketDG:
Issues like #142 and #260 will be properly handled during installation.


nvbn:
Thanks!

Issue #268, opened 2015-07-01
ImportError: No module named psutil


polybuildr:
Had to `sudo apt-get install python-psutil` to fix.

drblue:
Just a note, on Ubuntu 14.04.4 LTS the installation succeeded but trying to run `thefuck` resulted in error about `psutil` missing. Only got it working after I ran `sudo apt-get install python-psutil` as @polybuildr suggested.

Error-msg:
```
Traceback (most recent call last):
  File "/usr/local/bin/thefuck", line 7, in <module>
    from thefuck.main import main
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 10, in <module>
    from . import logs, types
  File "/usr/local/lib/python2.7/dist-packages/thefuck/types.py", line 6, in <module>
    from psutil import Process, TimeoutExpired
ImportError: No module named psutil
```

prasad-high-5:
I have installed psutils and trying to import psutil on SLES 12 platform, but still its shouting with no modules. Any suggestions please.


```
SLES_12:~ # rpm -qa | grep psutils
psutils-p17-1023.1.2.x86_64
SLES_12:~ #
SLES_12:~ # python -c 'import psutil'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named psutil
SLES_12:~ #
```

Liguo-Ma:
pip install psutil works well on my pc

Issue #273, opened 2015-07-03
AttributeError: 'module' object has no attribute 'PY2'


chatainsim:
Hello,

I've just installed thefuck through pip.
Add the .bashrc alias then source it.

But when I use fuck, I have this error : 

```
Traceback (most recent call last):
  File "/usr/local/bin/thefuck", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 129, in main
    command = get_command(settings, sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 73, in get_command
    if six.PY2:
AttributeError: 'module' object has no attribute 'PY2'
```

Do you know how to fix it ?

I'm using python Python 2.7.3.

Thanks

SanketDG:
`pip install six` should do the job.

chatainsim:
Thanks @SanketDG but six seems to be already installed : 
```
Requirement already satisfied (use --upgrade to upgrade): six in /usr/lib/python2.7/dist-packages
```
And when trying to upgrade : 
```
Collecting six
  Downloading six-1.9.0-py2.py3-none-any.whl
Installing collected packages: six
  Found existing installation: six 1.1.0
    DEPRECATION: Uninstalling a distutils installed project (six) has been deprecated and will be removed in a future version. This is due to the fact that uninstalling a distutils project will only partially uninstall the project.
    Can't uninstall 'six'. No files were found to uninstall.
Successfully installed six-1.1.0
```
But I still have the issue : 
```
Traceback (most recent call last):
  File "/usr/local/bin/thefuck", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 129, in main
    command = get_command(settings, sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 73, in get_command
    if six.PY2:
AttributeError: 'module' object has no attribute 'PY2'
```

SanketDG:
Probably upgrade pip and then try?
`pip install -U pip`
It's better if you install all these packages in a virtualenv.

chatainsim:
I've already upgrade pip to the latest version : 
```
pip 7.1.0 from /usr/local/lib/python2.7/dist-packages (python 2.7)
```

scorphus:
Hey, @chatainsim. Could you please give us a little bit more info?

 * What OS are you on?
[...]

PR #275, opened 2015-07-03
Update requirements.txt

Sticksman:
This should fix the requirements for the repo.


Sticksman:
fix #268 #273

SanketDG:
This is not needed. @Sticksman. `setup.py` handles all the dependencies. Only the test suite is needed.

PR #276, opened 2015-07-04
Add a tmux rule

mcarton:
Also fixes a spelling error in README.md, but I was too lazy to make another PR for that.


nvbn:
Thanks!

PR #277, opened 2015-07-04
Fix the pacman rule with `sudo`

mcarton:
Does not use @sudo_support as this does not place 'sudo' at the right position.
The `apt-get` rule might need to be fixed as well.


nvbn:
Thanks!

PR #279, opened 2015-07-04
fixed dependency problem, on python 3.4.0

mkreder:
Fixed dependency problem, on python 3.4.0 pathlib is included in the python distribution so it must not be included in the requirements for thefuck otherwise it wont run at startup.
This was necesary to make a package for Fedora
https://bugzilla.redhat.com/show_bug.cgi?id=1239148

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import sys
 
 VERSION = '1.46'
 
-deps = ['psutil', 'colorama', 'six']
+install_requires = ['psutil', 'colorama', 'six']
 
 if sys.version_info < (3,4):
     deps.append('pathlib')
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,8 +1,13 @@
 from setuptools import setup, find_packages
+import sys
 
 
 VERSION = '1.46'
 
+deps = ['psutil', 'colorama', 'six']
```
SanketDG: make this `install_requires`
and then do `install_requires=install_requires`

nvbn:
Thanks!

mkreder:
you're welcome!

PR #282, opened 2015-07-09
Add systemd's kind of error for the sudo rule

mcarton:
A complete error would be:

```
% systemctl daemon-reload
==== AUTHENTICATING FOR org.freedesktop.systemd1.reload-daemon ===
Authentication is required to reload the systemd state.
Authenticating as: martin
Password:
```


nvbn:
Thanks!

PR #283, opened 2015-07-09
Some fixes in REAME.md

mcarton:


scorphus:
Nice catch! Totally forgot it. Thanks!

mcarton:
You're welcome.

nvbn:
Thanks!

SanketDG:
Probably fix this [line](https://github.com/nvbn/thefuck/pull/283/files#diff-04c6e90faac2675aa89e2176d2eec7d8R204) too. It's confusing.

mcarton:
@SanketDG: indeed it is. I modeled it as the `apt_get` rule but that *or* now makes it wrong. Would something like:
```
* `pacman` &ndash; installs app with `pacman` if it is not installed (uses `yaourt` if available).
```
be better?

PR #288, opened 2015-07-14
fix(fish.get_aliases): do not include overridden aliases

### LINKED ISSUE #262: shells.Fish.get_aliases() should not include primary functions
Functions such as `cd` and `ls`. I'm on it and hope to get it fixed soon.

scorphus:
Fish Shell overrides some shell commands, such as `cd` and `ls` and therefore some rules fail to match. The following aliases are excluded by default:

 * cd
 * grep
 * ls
 * man
 * open

To change them, one can use the `TF_OVERRIDDEN_ALIASES` environment variable such as:

```fish
set TF_OVERRIDDEN_ALIASES 'cd,grep,ls'
```

Fix #262

### DIFF
--- a/tests/test_shells.py
+++ b/tests/test_shells.py
@@ -116,18 +116,34 @@ class TestFish(object):
     def Popen(self, mocker):
         mock = mocker.patch('thefuck.shells.Popen')
         mock.return_value.stdout.read.return_value = (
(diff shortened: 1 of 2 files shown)


nvbn:
Thanks!

PR #289, opened 2015-07-14
improve(rules): add a rule to recursively copy a directory

scorphus:
Add a new rule, `cp_dir`, that adds `-r` when you try to copy a directory:
```
» cp thefuck /tmp/
cp: thefuck is a directory (not copied).
» fuck
cp -r thefuck /tmp/ [enter/ctrl+c]
» ls /tmp/thefuck/
__init__.py
__init__.pyc
__pycache__
...
```


nvbn:
There's already exists `cp_omitting_directory` rule for doing similar stuff.
I added pattern from your rule to it.

scorphus:
Sure! Makes sense. I should have done that instead. Thanks!

PR #290, opened 2015-07-17
Resolve git aliases

mcarton:
This PR adds a new utility function to resolve git aliases for `git_*` rules.
It does not require to find, read or parse the `gitconfig` user file as it uses logging from `git` itself with the `GIT_TRACE` env variable.

It was inspired by the `git_diff_staged` rule by @scorphus which uses [some](https://github.com/nvbn/thefuck/blob/15e13d7c1a3b5733c84139e89b09d105bb70de93/tests/rules/test_git_diff_staged.py#L8-9) of [his aliases](https://github.com/scorphus/dotfilesetal/blob/26cea37aad58aed760547c7d578de4b82f9d1708/.gitconfig#L33-34).

### DIFF
--- a/tests/rules/test_git_diff_staged.py
+++ b/tests/rules/test_git_diff_staged.py
@@ -3,15 +3,13 @@ from thefuck.rules.git_diff_staged import match, get_new_command
 from tests.utils import Command
 
 
-@pytest.mark.parametrize('command', [
-    Command(script='git diff'),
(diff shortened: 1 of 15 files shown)


scorphus:
Nice job, @mcarton! Great contributions! Thank you! :+1:

nvbn:
Thanks!

Issue #287, opened 2015-07-13
OSX: "ImportError: No module named pathlib"


alexmwu:
Hi, I'm trying to install and I get this error:

```
$ fuck
Traceback (most recent call last):
  File "/usr/local/bin/thefuck", line 7, in <module>
    from thefuck.main import main
  File "/Library/Python/2.7/site-packages/thefuck/main.py", line 2, in <module>
    from pathlib import Path
ImportError: No module named pathlib
```

I tried updating setup by following the instructions in #218.
Running OSX 10.10.4, Python 2.7.6, and pip 7.1.0.
Any ideas?

AbdoulNdiaye:
I tried 
```sh

jonathan-villemot:
:+1:  Worked for me too.

paulagent:
Worked for me too. thx

alexmwu:
Oh well, that makes a lot of sense. It works now, thanks a lot!

Aaron3154518:
What is "pip" in sudo pip install pathlib?

m0caa:
> I tried
> 
> ```shell
> sudo pip install pathlib
> ```
> 
> and it works for me :)

THANK YOU

dopdahl16:
Should `pathlib` be added as a dependency in the documentation?

ShyamSundarAgwl:
I tried but it didn't worked.
Can you help me with any other solution?

PKQRAA:
> ```shell
> sudo pip install pathlib
> ```

ImportError: No module named utils.console

PR #291, opened 2015-07-17
Fix wheel dependencies for Python 2

petr-tichy:
Fix issues like #287 where `pathlib` wasn't installed on Python 2


nvbn:
Thanks!

PR #292, opened 2015-07-19
Add git_branch_delete rule

SimenB:
I have never written python before, so this is just copy-pasting and slashing. So if I missed some convention, or wrote the test in some wrong way, please tell me :smile: 

This is to be able to give a fuck in this situation:

image


nvbn:
Thanks!

PR #295, opened 2015-07-20
New git rules

mcarton:


nvbn:
Thanks!

nvbn:
Rule `git_push_force` is a bit dangerous, by default it'll be disabled.
Also I added rule `git_push_pull`, that pulls when push was rejected. I guess it'll better in most of rejections.

mcarton:
Ok, it seems a saner default behavior, I know `git_push_force` it seems a little dangerous (the rule would first add `-f`, I amended it afterwards to use `--force` instead to show users how bad it can be and then I `push -f` to edit the pull requests :)).

PR #299, opened 2015-07-21
Add `dnf history` error for the sudo rule

evverx:
```
$ dnf history
You don't have access to the history DB.
```


nvbn:
Thanks!

PR #304, opened 2015-07-21
Fix the `git_diff_staged` rule

mcarton:
The problem was:
```
% git add foo
% git diff foo
% fuck
git diff foo --staged [enter/ctrl+c]
fatal: bad flag '--staged' used after filename
```


nvbn:
Thanks!

Issue #253, opened 2015-06-11
Git push on new branch


eranimo:
When running `git push` on a new branch

```
fatal: The current branch update-project-availability has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin [branch_name]
```

It would be nice if I can just say `fuck` and it will run that command.

Currently it gives `no fucks given`

bugaevc:
Works fine for me on the latest version (and the rule responsible for that is [git_push.py](https://github.com/nvbn/thefuck/blob/master/thefuck/rules/git_push.py)).

scorphus:
Same as @bugaevc's here. @eranimo, perhaps your Git is a bit too old that its error message is _incompatible_ with the `git_push` rule. What does `git --version` tell you?

elexx:
I had the same problem. It turns out I used githubs `hub` command on top of git. As github suggests I aliased git to hub `alias git=hub`. It seems this confuses thefuck.

bugaevc:
Can we add `@alias_support` decorator that would resolve aliases before the command analysis and try to restore them back afterwards? /cc @nvbn

eranimo:
I'm not aliasing git, and it's still not working. I'm on git version 2.3.2 btw

jab:
Is there a workaround for this that doesn't require removing the git=hub alias?

chrimifi:
I'm seeing this issue when I use the fish shell, but not bash. I think it's because the idiomatic way to do aliases in fish is to create a function (and if you use alias the shell will do this for you), and thefuck expects something tighter? I haven't had the chance to investigate it myself yet.

scorphus:
I really can't reproduce it. What are your Git and Fish versions, @chrimifi?

scorphus:
BTW this is what I get:
```fish
⋊> ~/W/barebones on works-for-me ● thefuck git push
git push --set-upstream origin works-for-me [enter/↑/↓/ctrl+c]
git push --set-upstream origin works-for-me
⋊> ~/W/barebones on works-for-me ●
```

chrimifi:
Fish 2.2.0
git 2.8.0
hub 2.2.3

And if I type `functions git`, I get:
function git
	command hub $argv
end

chrimifi:
Here's what I get trying your commands:
```
finch@Lindblum $ git push                                                                                                                   12:44 31-03-2016
fatal: The current branch frobulate has no upstream branch.
[...]

PR #306, opened 2015-07-21
Support GitHub's hub command

mcarton:
GitHub's `hub` command is [supposed to be used with `alias git=hub`](https://github.com/github/hub#aliasing).

See: https://github.com/nvbn/thefuck/issues/253#issuecomment-111799922


nvbn:
Thanks!

PR #310, opened 2015-07-24
Make a `sudo` pattern more general

mcarton:
```
% mount -o uid=martin /dev/sdc1 mnt
mount: only root can use "--options" option
```


nvbn:
Thanks!

PR #317, opened 2015-07-26
fix setup.py version checking

SanketDG:
There was a small issue with the `setup.py` that I had implemented in #267. `version_info` returns a five member `tuple`, instead `version_info[:2]` has to be done to get the required version.


nvbn:
Thanks!

PR #321, opened 2015-07-28
Force the travis image to track the master branch

mcarton:
It is currently showing that the build fails, but this only concerns a test branch.


nvbn:
Thanks!

PR #333, opened 2015-08-13
basic support for the hdfs dfs <command> when the command misses the …

mlk:
…dash

`hdfs dfs ls` will be replaced with `hdfs dfs -ls`.


README.md
```diff
@@ -192,6 +192,7 @@ using the matched rule and runs it. Rules enabled by default are as follows:
 * `tsuru_not_command` &ndash; fixes wrong tsuru commands like `tsuru shell`;
 * `tmux` &ndash; fixes `tmux` commands;
 * `whois` &ndash; fixes `whois` command.
+* `hdfs dfs` &ndash; Add the missing dash to the command.
```
brittneybrinsfield: Dash

mlk: Dash? Have I missed something?

mcarton: You actually have named your rule `hadoop_dfs_missing_dash` and that list was in alphabetic order.

mlk: Ah got ya. OK will change and resubmit.

tests/rules/test_hadoop_dfs_missing_dash.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+from thefuck.rules.hadoop_dfs_missing_dash import match, get_new_command
+from tests.utils import Command
+
+
+@pytest.mark.parametrize('command', [
+    Command(script='./bin/hdfs dfs ls', stderr='ls: Unknown command\nDid you mean -ls?  This command begins with a dash.'),
```
mcarton: The command gives a really nice error message: it has the faulty command and the fix for it. Maybe you could make your rule more generic? You can get inspiration from the [*git_not_command*](https://github.com/nvbn/thefuck/blob/master/thefuck/rules/git_not_command.py) rule.

mlk: That makes sense. 

Should it then ignore the command and just look for the error message  ": Unknown command" / "Did you mean ?"?

mcarton: I meant more generic in the sense that there are probably other subcommands than '-ls' no? (I don't know `hdfs`)
Anyway yes, it would probably be good to fix any command whose output is of the form "foo: Unknown command\nDid you mean bar?"

mlk: The previous solution handled any hdfs dfs <blah> already as it blinded adds a "-" to the second param.

But I've now pushed a version that ignores the command and just looks at the error message.

mcarton: Oh indeed sorry, I got confused by all those "fs/ls" :$

README.md
```diff
@@ -162,6 +162,7 @@ using the matched rule and runs it. Rules enabled by default are as follows:
 * `go_run` &ndash; appends `.go` extension when compiling/running Go programs
 * `grep_recursive` &ndash; adds `-r` when you trying to `grep` directory;
 * `gulp_not_task` &ndash; fixes misspelled gulp tasks;
+* `hadoop_dfs_missing_dash` &ndash; Add the missing dash to the command.
```
mcarton: Last comment and I stop annoying you for today: all the other items start with a verb conjugated at the 3rd person and with a lower case letter, and all but the last items end with a ';' not a '.'.

mlk: No worries. Really it is no longer a fix for hdfs either.

nvbn:
Thanks!

PR #342, opened 2015-08-20
Issue: 322 Runs vagrant up when required

mlk:
Starts up a vagrant instance if the user runs a command (such as vagrant ssh) when the instance is not currently running.


nvbn:
Thanks!

PR #343, opened 2015-08-21
hdfs -rm -r /directory and hdfs -mkdir -p /directory/sub support

mlk:
This expands the rm_dir and mkdir_p rules to also work with hdfs commands.


nvbn:
Thanks!

Issue #334, opened 2015-08-13
"history" rule runs very slow when bash history is huge


felixonmars:
thefuck became very slow since a few versions ago. After specifying `THEFUCK_DEBUG=true` I have found the problem:

```
...
DEBUG: Trying rule: history; took: 0:00:07.860372
...

I have 100k+ lines in ~/.bash_history, and that should be the reason. Is there any way to speed up the rule, or maybe I have to clean up bash history regularly?

Thanks!

kloetzl:
Even with my history being just 1600 lines, the rule took 0.25 seconds. That might not sound like much, but its a noticeable delay. I think, executing `fuck` should really *feel* instantaneous i.e. the result must not take longer than 50ms.

hcgraf:
Suggestion: try to use the version for python 3, the history rule seems to be MUCH faster…

pip uninstall thefuck
pip3 install thefuck

felixonmars:
@hcgraf Thanks for the tip, but I'm already using Python 3 :)

PR #345, opened 2015-08-23
Fix slowness problems

mcarton:
While searching for a solution for #334 I discovered that the `@memoize` annotation is not applied to the right place and so the shell was always executed twice to get the aliases list.
On my system with zsh:
```shell
# before:
% time thefuck test
No fucks given
thefuck test  0.87s user 0.10s system 96% cpu 1.003 total
# after:
% time thefuck test           
No fucks given
thefuck test  0.50s user 0.07s system 96% cpu 0.588 total
```

Now that this is fixed, the `history` rule is still the longest one, but this has become reasonable (~40ms with a 10k lines history file on my system).

### DIFF
--- a/tests/rules/test_mercurial.py
+++ b/tests/rules/test_mercurial.py
@@ -2,7 +2,7 @@ import pytest
 
 from tests.utils import Command
 from thefuck.rules.mercurial import (
-    extract_possisiblities, match, get_new_command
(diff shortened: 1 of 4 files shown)


nvbn:
Thanks!

scorphus:
Thanks, Martin! :+1:

Issue #35, opened 2015-04-18
add maven support


sonenko:
It would be great to support maven commands. See example below. (should be `mvn clean install`, see last lines of the stack trace)
```
$ mvn clean insta
```
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option PermSize=128m; support was removed in 8.0
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
[INFO] Scanning for projects...
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building Camel :: Example :: Twitter WebSocket 2.15.1
[INFO] ------------------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 0.536 s
[INFO] Finished at: 2015-04-18T23:15:13+03:00
[INFO] Final Memory: 22M/981M
[INFO] ------------------------------------------------------------------------
[ERROR] Unknown lifecycle phase "insta". You must specify a valid lifecycle phase or a goal in the format <plugin-prefix>:<goal> or <plugin-group-id>:<plugin-artifact-id>[:<plugin-version>]:<goal>. Available lifecycle phases are: validate, initialize, generate-sources, process-sources, generate-resources, process-resources, compile, process-classes, generate-test-sources, process-test-sources, generate-test-resources, process-test-resources, test-compile, process-test-classes, test, prepare-package, package, pre-integration-test, integration-test, post-integration-test, verify, install, deploy, pre-clean, clean, post-clean, pre-site, site, post-site, site-deploy. -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/LifecyclePhaseNotFoundException
```
$ fuck 
```
No fuck given

MarekAG:
I think that it's probably impossible to make TheFuck work with mvn.
mvn stack trace does not give us proper command directly and mvn errors aren't as repeatable as e.g. when we forget to put "sudo" before command.

mlk:
@MarekAG why do you think that? 

[...]

PR #351, opened 2015-08-26
#35 mvn support

mlk:
Includes two new rules:

*  `mvn_no_command` will automatically turn `mvn` into `mvn clean package` as that is what you really want.
* `mvn_unknown_lifecycle_phase` will run a fuzzy search over the returned available life cycle stages.


nvbn:
Thanks!

PR #354, opened 2015-09-01
Some README fixes

mcarton:
It's either *misspelled* (US) or *misspelt* (UK), I went for "misspelled" as it was already used in the README.
Also punctuation.


nvbn:
Thanks!

PR #373, opened 2015-09-12
Add j, k key for arrow action at read_actions

grammaright:
Hello nvbn.

I think that `j` and `k` key are known to many shell users as navigator keys. (such as vim, less, etc..) So i added `j` `k` for move ↑ or ↓.


nvbn:
Thanks!

PR #374, opened 2015-10-02
Test on python 3.5

mcarton:
Python 3.5 has been out and available on travis for some time now, but we were blocked by a *py.test* bug. Now that it's fixed, we should test on Python 3.5 as well.


nvbn:
Thanks!

PR #375, opened 2015-10-02
rm -f $eval_script

jab:
Without this change, users who have "rm" aliased to "rm -i"
have to confirm removal of the file after running fuck. This
change allows such users to run fuck without having to do
the superfluous rm confirmation.


jab:
Alternatively, changing `rm` to `command rm` or `/bin/rm` would bypass any aliases the user set up and run the `rm` command directly. Do you have a preference?

jab:
I just changed it to /bin/rm for now since I think that's cleanest.

nvbn:
Thanks!

Issue #89, opened 2015-04-22
feature request: replace history with corrected command.


viralex:
It would be a nice feature to correct the command and the history.
I would also like an option to not add {fuck,thefuck} to the history.

Dugucloud:
Cool, but maybe it is decided by the shell?

nwinkler:
To avoid adding the `fuck` commands to history, you can use the following in Bash:

    export HISTIGNORE=' *'

Then any command prefixed with a space will not be added to history. So instead of typing `fuck`, you would need to type ` fuck` (with a space as the first character), and it will not be added to the shell's history.

viralex:
I use zsh, and It seems that the HISTIGNORE trick it's not working. (but It's not so important I've already set to not save duplicates). 

I think thefuck could execute "history" command to update the history. I've not tested it but man says:

 history change newValue ?event?
              Replaces the value recorded for  an  event  with  newValue.   Event
              specifies  the  event to replace, and defaults to the current event
              (not event -1).  This command is intended for use in commands  that
              implement new forms of history substitution and wish to replace the
              current event (which invokes the  substitution)  with  the  command
              created through substitution. 

is it cross-shell compatible? :\

jmeosbn:
I was just about to suggest this. :+1: 

Avoiding adding `fuck` commands can at least be configured per shell, whereas adding the modified commands cannot.  Ideally the original command would also be left intact (in case thefuck was wrong).  Removing the `fuck` command from within `thefuck` could be difficult without some hack.

While `fc` can work differently in different shells, the example below works great in `bash` - removing the `fc` command from history while keeping the original and modified commands in the history. In `zsh` the example works the same, except the `fc` command remains in history.

```sh
$ ls -l /usr
$ fc -e- /usr=/etc
```

History in `bash` afterward:

```
ls -l /usr
ls -l /etc
```

nvbn:
Implemented for zsh and bash in 1.35+

scorphus:
As soon as I have some spare time I can dig this one and implement it for Fish.

scorphus:
The solution to this issue Fish-wise depends on fish-shell/fish-shell#825 available in the upcoming release, [Fish 2.2.0](/fish-shell/fish-shell/milestones/fish 2.2.0). I'll keep an eye on that and update the `fuck` Fish function as soon as that's released.

nvbn:
So now it works.

PR #384, opened 2015-10-16
Improve the Fish Shell function making it faster

scorphus:
It works now with no temp file involved, which makes it a lot faster. Also, `history --merge`, although only supported on Fish Shell 2.2+, merges the corrected entry back into history. Neat!

Ref #89


nvbn:
Thanks!

Issue #383, opened 2015-10-15
Add support for the XDG Base Directory Specification


SShrike:
You can find the specification [here](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

This [blog post](https://ploum.net/207-modify-your-application-to-use-xdg-folders/) lists the advantages of using the XDG Base Directory Specification as follows:
- `$HOME` is a lot less cluttered.
- Backups are a lot more safer and easier (you know that creating a backup of your `$XDG_DATA_HOME` along with your files is enough).
- A lot easier to reset a default configuration if you want/need it (and without any risk of loosing information).
- Avoid some strange bugs that happen because you had a old version of some configuration file.
- It is a lot more flexible and portable because no paths are hard-coded. You can use the XDG library that does the job for you _or_, if you don’t want the dependency, implementing the XDG specification is only a few lines of code.

SShrike:
#385 has been merged.

PR #385, opened 2015-10-16
 Add support for the XDG Base Directory Specification #383

mcarton:
This is a redo of one of my oldest PR here (#160) that I got bored updating for nothing a while ago.
I'm obviously not the only one interested with that:
- #160 ;
- #383 *(explains well why this is useful)* ;
- https://github.com/andrewmichaud/thefuck/commit/a90c80ba534363430cb0ba3cafd2e9a288aa2c7a.

This is a non-breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ Bundled, but not enabled by default:
 ## Creating your own rules
 
 For adding your own rule you should create `your-rule-name.py`
-in `~/.thefuck/rules`. The rule should contain two functions:
+in `~/.config/thefuck/rules`. The rule should contain two functions:
 
 ```python
 match(command: Command) -> bool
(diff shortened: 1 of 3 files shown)


scorphus:
:+1:

How about moving `˜.thefuck` to `$XDG_CONFIG_HOME/thefuck` if the former exists?

mcarton:
Maybe, but that would break the conf of people who currently back up
`~/.thefuck` or have `~/.thefuck` be a symlink to `~/.config/thefuck` to
be able to back up `./config` (such as me, so I'm a little biassed about
that :)).
Other tools that have adopted `$XDG_CONFIG_HOME` seem to keep the
historical directory for compatibility.

SShrike:
I'm on the same page as @mcarton, moving a user's existing configuration is _not_ a good idea.

scorphus:
Yeah, makes sense. I didn't think much about the complications it may raise.

nvbn:
Thanks!
Currently `~/.thefuck` have a greater priority, then `~/.config/thefuck`, so it shouldn't break compatibility, even with old config's backups.

PR #388, opened 2015-10-17
Fix rules.brew_install tests on Mac

scorphus:
As these tests are skipped on Travis and for most of other devs, they were failing unnoticeably.


nvbn:
Thanks!

PR #389, opened 2015-10-18
Fix `rules.touch` tests

scorphus:
Move them to `rules` sub-directory and import `shells` instead of `and_` which in turn triggers the `generic_shell` fixture fixing the tests on Fish Shell.


nvbn:
Thanks!

PR #390, opened 2015-10-22
add THEFUCK_HISTORY_LIMIT, my machine is so slow

lovedboy:
when my history commands is greater than 10000 lines, it will cost me one second to run fuck command. So i wish i can control how many history command the fuck will scan in order to match the closest command.


lovedboy:
I have no idea why  checks  failed.

PR #392, opened 2015-10-28
Better splitting

mcarton:
Use `shlex` to parse the command, this correctly handles quotes.
For example, rules `dirty_un{tar,zip}` did not handle filenames with spaces (be them escaped or quoted).
Also the splitting is now done once and cached, instead of in almost every rule.

This won't magically fix all rules as some rules (most of them I quick-fixed in e71a3e0) join the arguments incorrectly (but it's likely they were already broken in those cases).


nvbn:
Thanks!

PR #397, opened 2015-10-30
Use --force-with-lease instead of --force for git push

janek-warchol:
--force flag can be very dangerous, because it unconditionally
overwrites remote branch - if someone pushed new commits to the remote
repo after you last fetched/pulled, and you do push --force, you will
overwrite his commits without even knowing that you did that.  Using
--force-with-lease is much safer because it only overwrites remote
branch when it points to the same commit that you think it points to.

Read more:
https://developer.atlassian.com/blog/2015/04/force-with-lease/

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Enabled by default only on specific platforms:
 
 Bundled, but not enabled by default:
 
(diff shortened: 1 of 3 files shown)


nvbn:
Thanks!

Issue #404, opened 2015-11-15
Ugly warning while selecting command


Kwaadpepper:
```The Fuck 3.2 using Python 2.7.10```
```Kubuntu 15.10```
In **.bashrc** ```alias fuck='eval "$(eval $(thefuck $(fc -ln -1)))"'```
In **.zshrc** ```eval $(thefuck --alias)```
So with bash, and zsh

I tried thefuck with the **ubuntu  repo version**, which happened to be fixed recently, i tried with **the last updated version of thefuck**

In the first example i ran: ```sudo aptget update``` then  ```fuck``` then type <kbd>↑</kbd> (up arow) on both **zsh** and **bash**

    $ fuck
    sudo apt-get update [enter/↑/↓/ctrl+c]/usr/local/lib/python2.7/dist-packages/thefuck/rules/history.py:24: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal and line.split(' ')[0] in executables]
    sudo plget update [enter/↑/↓/ctrl+c]
    Aborted

The debug version http://paste.ubuntu.com/13282061/

It seems to happen on any command, i tried with ```sulb``` instead of ```subl```

mcarton:
That's yet another problem with Unicode and Python 2. If you can, use The Fuck with Python 3 (eg. install it with `pip3 install thefuck`), the problem will likely be gone. @scorphus: if you still want to fix that :smile:

BTW: you can use `eval "$(thefuck --alias)"` within **.bachrc** too, that way if the alias ever get updated, you'd automatically use the new one.

Kwaadpepper:
I just got ```python3-pip``` and done ```pip3 install thefuck```
I got errors on install which i give you in these link (second is the debug file)
http://paste.ubuntu.com/13282680/
http://paste.ubuntu.com/13282685/

Running thefuck gives

    Traceback (most recent call last):
      File "/usr/local/bin/thefuck", line 7, in <module>
    from thefuck.main import main
      File "/usr/local/lib/python3.5/dist-packages/thefuck/main.py", line 6, in <module>
    from . import logs, types, shells
      File "/usr/local/lib/python3.5/dist-packages/thefuck/types.py", line 6, in <module>
    from psutil import Process, TimeoutExpired
    ImportError: No module named 'psutil'

It was missing psutil and decorator, which i installed.
And finally i got this **thefuck** error

http://paste.ubuntu.com/13282783/

[...]

PR #406, opened 2015-11-15
Fix cache problem when going from Python 2 to 3

mcarton:
See https://github.com/nvbn/thefuck/issues/404#issuecomment-156816429.


nvbn:
Thanks!

PR #394, opened 2015-10-29
add THEFUCK_HISTORY_LIMIT, my machine is so slow

lovedboy:
when my history commands is greater than 10000 lines, it will cost me one second to run fuck command. So i wish i can control how many history command the fuck will scan in order to match the closest command.

PR #409, opened 2015-11-19
#394 history limit

nvbn:


lovedboy:
@nvbn  `history_limit &ndash numeric value of how many history commands will be scanned, like 2000;` need be fixed.

nvbn:
@lovedboy thanks)

PR #413, opened 2015-11-29
Delete NULL bytes from command.script_parts[0]

ilovezfs:
The byte sequence in command.script_parts[0] can contain one or more
'\x00' which will cause os.stat(path) to throw this error:

        TypeError: must be encoded string without NULL bytes, not str


nvbn:
The code isn't working, rule and tests fails with:
```
 TypeError: translate() takes exactly one argument (2 given)
```

ilovezfs:
ug http://stackoverflow.com/questions/23175809/typeerror-translate-takes-one-argument-2-given-python

Issue #378, opened 2015-10-08
Usage freezes terminal in Windows


pingzing:
Invoking `thefuck` in any way causes my terminal to lock up, using any shell--Powershell (via ConEmu) vanilla Powershell, or msysgit's git bash.

For example, `thefuck git brnch` outputs a slightly malformed suggestion of "git branch" and then stops accepting user input entirely.

I'm running Windows 10 and 32-bit Python 3.5.0.

Powershell
![fuck](https://cloud.githubusercontent.com/assets/5133649/10382102/3043a6f8-6e29-11e5-9f2d-75436bd85696.png)

Git bash
![fuck](https://cloud.githubusercontent.com/assets/5133649/10382113/462c12f2-6e29-11e5-8bb6-f31b58d00d1e.png)

hchokshi:
I'm also experiencing the same issue using Powershell and the msysgit bundled with Github for Desktop. Shows the correct command suggestion, but doesn't take any input after that.

I'm also seeing \u2191 and \u2198 in the console like you are, here. I doubt this is intentional. 

These issues are experienced directly opening Powershell / bash and using them through cmder.

TAGC:
I'm experiencing this as well (through both Cygwin and Git Bash for Windows).

scharney:
Sorry to necro-post, but I'm having this same issue, and I'm not sure what's causing it. I'd love to be able to get it to work if anyone has suggestions! I'm using git bash on Windows 11.

PR #414, opened 2015-11-30
Fix getch on windows

pakrym:
Fixes https://github.com/nvbn/thefuck/issues/378

### DIFF
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -2,6 +2,8 @@ from argparse import ArgumentParser
 from warnings import warn
 from pprint import pformat
 import sys
+import win_unicode_console
+win_unicode_console.enable() #https://github.com/tartley/colorama/issues/32
 import colorama
 from . import logs, types, shells
 from .conf import settings
--- a/thefuck/ui.py
+++ b/thefuck/ui.py
@@ -6,7 +6,19 @@ from .exceptions import NoRuleMatched
 from . import logs
 
 try:
-    from msvcrt import getch
+    import msvcrt
+    def getch():
+        ch = msvcrt.getch()
+        if ch in (b'\x00', b'\xe0'):  # arrow or function key prefix?
+            ch = msvcrt.getch()  # second call returns the actual key code
+
+        if ch == b'\x03':
+            raise KeyboardInterrupt
+        if ch == b'H':
+            return 'k'
+        if ch == b'P':
+            return 'j'
+        return ch.decode(sys.stdout.encoding)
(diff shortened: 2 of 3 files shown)


thefuck/main.py
```diff
@@ -2,6 +2,8 @@
 from warnings import warn
 from pprint import pformat
 import sys
+import win_unicode_console
+win_unicode_console.enable() #https://github.com/tartley/colorama/issues/32
```
nvbn: Is the issue related here?

thefuck/ui.py
```diff
@@ -6,7 +6,19 @@
...
-    from msvcrt import getch
+    import msvcrt
+    def getch():
+        ch = msvcrt.getch()
+        if ch in (b'\x00', b'\xe0'):  # arrow or function key prefix?
+            ch = msvcrt.getch()  # second call returns the actual key code
+
+        if ch == b'\x03':
+            raise KeyboardInterrupt
+        if ch == b'H':
```
nvbn: `msvcrt.getch` returns `H` for `↑` and `P` for `↓`?

nvbn:
I've made a few changes, please check branch `pakrym-master`.

thefuck/ui.py
pakrym: Yes. It returns two chars one \xe0 and one of those.

thefuck/main.py
pakrym: Yes, you need to init `win_unicode_console` before importing `colorama`

pakrym:
You moved `win_unicode_console.enable()` call after importing `colorama` they fixed the issue (https://github.com/tartley/colorama/issues/32) but this change is not in `colorama-0.3.3` which `thefuck` uses.

nvbn:
@pakrym they said:
> Since win_unicode_console replaces the streams in order to support Unicode, win_unicode_console.enable has to be called before colorama.init so everything works as expected.

So `win_unicode_console.enable` call is before `colorama.init`.

Also I've updated code in branch `pakrym-master`, please check is it works on windows correctly =)

pakrym:
I meant other part of it:
> As of colorama v0.3.3, there is an early binding issue (https://github.com/tartley/colorama/issues/32), so win_unicode_console.enable has to be called even before importing colorama

nvbn:
@pakrym ok, now it calls `win_unicode_console.enable` before import.

Is it work ok?

nvbn:
Checked by myself, seems working.

Issue #415, opened 2015-12-10
Simple prepending of "sudo" is not always helps


krasovsky-evgeny:
```bash
user@machine:/$ mkdir /var/www
mkdir: cannot create directory ‘/var/www’: Permission denied
user@machine:/$ fuck
sudo mkdir /var/www [enter/↑/↓/ctrl+c]
```
root is owner of /var/www

```bash
user@machine:/$ touch /var/www/1/2/3
touch: cannot touch ‘/var/www/1/2/3’: No such file or directory
user@machine:/$ fuck
mkdir -p /var/www/1/2 && touch /var/www/1/2/3 [enter/↑/↓/ctrl+c]
```

the command is ok, we have permission error
```bash
mkdir: cannot create directory ‘/var/www/1’: Permission denied
user@machine:/$ fuck
sudo mkdir -p /var/www/1/2 && touch /var/www/1/2/3 [enter/↑/↓/ctrl+c]
touch: cannot touch ‘/var/www/1/2/3’: Permission denied
```

permission error related to touch, but accroding to common pattern ("sudo + command") we have double sudo

```bash
user@machine:/$ fuck
sudo sudo mkdir -p /var/www/1/2 && touch /var/www/1/2/3 [enter/↑/↓/ctrl+c]
touch: cannot touch ‘/var/www/1/2/3’: Permission denied
```

makalaaneesh:
@nvbn Hey, as of now, you've just prevented the double sudo. If I'm not wrong, the permission error for the second command (after the '&&') will still persist. 
Can i work on this issue?

nvbn:
@makalaaneesh sure!

nvbn:
Can I close this issue?

krasovsky-evgeny:
Yes, I think

makalaaneesh:
Yup!

PR #418, opened 2015-12-23
sudo sh execute for && in commands - preventing double sudo

makalaaneesh:
Followed the same procedure in handling this as in handling ```'>' in command.script```
@nvbn This is in reference to #415


nvbn:
Thanks!

PR #423, opened 2016-01-05
Add Windows error message support to cd_mkdir rule

MattKotsenas:
Add the Windows error message 'the system cannot find the path specified'
to the list of recognized messages for cd_mkdir.


MattKotsenas:
I see I failed the build for Python 2.7, but not 3.x. I'm still new to Python so any help is greatly appreciated. Thanks!

mcarton:
It should be OK in python 2.7 too and the error seems TIMEOUT-related in some unrelated part of the code. Re-run the tests just in case (don't know if travis will let you, but a `git commit --amend && git push -f` should do it).

nvbn:
Thanks!

PR #424, opened 2016-01-06
adding dnf install rule

IuryAlves:


thefuck/rules/dnf_install.py
```diff
@@ -0,0 +1,22 @@
...
+
+def get_packages(command):
+    splitted_command = command.script.split()
+    if splitted_command[0] == 'sudo':
+        return splitted_command[3:]
+    return splitted_command[2:]
+
+
+def match(command):
+    typos = ('dfn install', 'dnf istall', 'dfn istall')
```
mcarton: This is not very generic:
  * The case `dnf` vs. `dfn` vs. any other similar typo is already taken care of by the `no_command` rule;
  * The case of `istall` vs. `install` should be taken care of with `utils.get_closest` and a list of possible candidates (I guess there is more to `dnf` than just an `install` subcommand).

thefuck/rules/dnf_install.py
```diff
@@ -0,0 +1,22 @@
+
+
+def get_packages(command):
+    splitted_command = command.script.split()
+    if splitted_command[0] == 'sudo':
```
mcarton: You can handle `sudo` with the `@sudo_support` annotation.

README.md
```diff
@@ -216,6 +216,7 @@ Enabled by default only on specific platforms:
 * `brew_upgrade` &ndash; appends `--all` to `brew upgrade` as per Homebrew's new behaviour;
 * `pacman` &ndash; installs app with `pacman` if it is not installed (uses `yaourt` if available);
 * `pacman_not_found` &ndash; fixes package name with `pacman` or `yaourt`.
+* `dnf_install` &ndash; fixes `dnf install` command
```
mcarton: The rule lists are usually alphabetized.

thefuck/rules/dnf_install.py
IuryAlves: * When a type ```dfn install``` fuck only gives me ```idn_install``` and ```df install```.
* Sorry, i didn't notice ```utils.get_closest```. I will fix it. And yes. There is much more for dnf command, but i wanted to start with a short commit, since this is my first one for this project.

IuryAlves:
@mcarton What you think of using ```difflib``` to find possible typos?

Something like

```
from difflib import SequenceMatcher
typo = "istall" 
correct_subcommand = "install"

diff = SequenceMatcher(typo, correct_subcommand)
dif.ratio()  # something greater than 0.8

**Comment by mcarton** [comment]
@IuryAlves We already have a helper [`utils.replace_command`](https://github.com/nvbn/thefuck/blob/acd0b3e0244d28ef977653f8d4b9a8b0b277237f/thefuck/utils.py#L137) for that that uses `difflib` under the hood. In your case it could be something like that:
```python
replace_command(command, command.script_parts[1], ['install', 'reinstall', 'search', …])
```
Also prefer `command.script_parts` to `command.script.split()`, the former is shell-aware.

Issue #425, opened 2016-01-07
unexpected end of regular expression with `git u[`


robkorv:
* The Fuck 3.2 using Python 2.7.6
* /bin/bash, 4.3.11(1)-release
* Ubuntu 14.04
* [debug.txt](https://github.com/nvbn/thefuck/files/81297/debug.txt)

```python
Traceback (most recent call last):
  File "/usr/local/bin/thefuck", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 77, in main
    fix_command()
  File "/usr/local/lib/python2.7/dist-packages/thefuck/main.py", line 28, in fix_command
    selected_command = select_command(corrected_commands)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/ui.py", line 97, in select_command
    selector = CommandSelector(corrected_commands)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/ui.py", line 60, in __init__
    self._commands = [next(self._commands_gen)]
  File "/usr/local/lib/python2.7/dist-packages/thefuck/corrector.py", line 43, in organize_commands
    first_command = next(corrected_commands)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/corrector.py", line 74, in <genexpr>
    for corrected in rule.get_corrected_commands(command))
  File "/usr/local/lib/python2.7/dist-packages/thefuck/types.py", line 234, in get_corrected_commands
    new_commands = compatibility_call(self.get_new_command, command)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/utils.py", line 237, in compatibility_call
    return fn(*args)
  File "<decorator-gen-58>", line 2, in get_new_command
  File "/usr/local/lib/python2.7/dist-packages/thefuck/specific/git.py", line 31, in git_support
    return fn(command)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/rules/git_not_command.py", line 17, in get_new_command
    return replace_command(command, broken_cmd, matched)
  File "/usr/local/lib/python2.7/dist-packages/thefuck/utils.py", line 137, in replace_command
    for new_cmd in new_cmds]
  File "/usr/local/lib/python2.7/dist-packages/thefuck/utils.py", line 110, in replace_argument
    script, count=1)
  File "/usr/lib/python2.7/re.py", line 151, in sub
    return _compile(pattern, flags).sub(repl, string, count)
[...]

PR #427, opened 2016-01-07
#425 command had to be re escaped

makalaaneesh:
The ```'['``` in the broken command was interpreted as a python-re **special character**


nvbn:
Thanks!

robkorv:
tnx @makalaaneesh, it's fixed. :cake:

PR #443, opened 2016-01-24
new pattern in sudo rule

aureooms:
`pacman-key --refresh-keys` outputs the following when not run as root

```
==> ERROR: pacman-key needs to be run as root for this operation.
```


nvbn:
Thanks!

PR #447, opened 2016-01-26
Changed dbm to anydbm for better not-unix support

kizetsu:
I changed the dbm import to anydbm to fix the issue on windows. At my computer it is running very well. It have to be testet on Linux and OSX too.
Like it's written in the python docs (https://docs.python.org/2/library/anydbm.html) the anydbm is a generic interface of the dbm database that is supported from the most systems. So I hope that it works well on linux and osx too.
I created a new branch so you can get the changes to test first, if you want. Maybe you could test one of the systems, because I actually haven't the possibilities.

TODO:
- Test on Linux
- Test on OSX

### DIFF
--- a/thefuck/utils.py
+++ b/thefuck/utils.py
@@ -1,4 +1,4 @@
-import dbm
+import anydbm
 import os
 import pickle
 import pkg_resources
@@ -16,10 +16,10 @@ from warnings import warn
 
 DEVNULL = open(os.devnull, 'w')
 
-shelve_open_errors = (dbm.error, )
-if six.PY2:
-    import gdbm
-    shelve_open_errors += (gdbm.error, )
+shelve_open_errors = (anydbm.error, )
+#if six.PY2:
+#    import gdbm
+#    shelve_open_errors += (gdbm.error, )
 
 
 def memoize(fn):


mcarton:
Unfortunately *anydbm* has been renamed *dbm* in Python3 (Python 2’s *dbm* has been renamed *dbm.ndbm*) which is why your PR fails on all Python3 versions. You can use [that trick](https://github.com/nvbn/thefuck/blob/3ad953001d829555613f060d1315ee53f8cff2d6/thefuck/rules/cd_correction.py#L40-L43) to fix it.
Travis tests on Linux by the way.

thefuck/utils.py
```diff
@@ -16,10 +16,10 @@
...
 DEVNULL = open(os.devnull, 'w')
 
-shelve_open_errors = (dbm.error, )
-if six.PY2:
-    import gdbm
-    shelve_open_errors += (gdbm.error, )
+shelve_open_errors = (anydbm.error, )
+#if six.PY2:
+#    import gdbm
+#    shelve_open_errors += (gdbm.error, )
```
scorphus: Add it at this line instead

kizetsu:
Thank you for the hint. I completely didn't see that it's renamed in Python3.
I will fix it as soon I get the time and let you know when it's done.

scorphus:
Good! You can amend the one commit in this pull request and do a forced push to the same branch – `hotfix/wp-windows-dbm-issue` – and GitHub will magically bring it all together.

scorphus:
@kizetsu: ping, any updates on this? Thanks.

Issue #444, opened 2016-01-24
Problems at installation


lkmnds:
When installing thefuck from pip(everything ok after that), then I run "fuck" and it shown me an error, saying psutil is not found(ImportError), after I've installed psutil from pip it worked, so because of that, thefuck still has psutil as a dependency, but the README.md says that it doesn't.

vkotovv:
I've created a [pull request](#456) for that.

nvbn:
@lkmnds, can you post output of `pip --version`?

lkmnds:
pip 1.5.6 from /usr/lib/python2.7/dist-packages (python 2.7)

got pip from apt-get in a fresh debian install.

PLNech:
I confirm this issue, I did need to install a few more python packages than listed in `requirements.txt`:
- psutil
- colorama
- decorator

Should I open a separate issue regarding those?

> uname -a
Darwin MacBook-Pro.local 15.3.0 Darwin Kernel Version 15.3.0: Thu Dec 10 18:40:58 PST 2015; root:xnu-3248.30.4~1/RELEASE_X86_64 x86_64


> pip --version
pip 8.0.3

nvbn:
This kind of problem was caused by outdated pip, but now we check for pip version in `setup.py`.

PR #456, opened 2016-02-06
Added missing psutil dependency for install command

vkotovv:
See #444 for details. Should mention psutil in Readme


nvbn:
`psutil` already mentioned in `setup.py` as `install_requires` - https://github.com/nvbn/thefuck/blob/master/setup.py#L25.

vkotovv:
So, #444 could not really happen?

nvbn:
It happens with some old versions of pip, I guess it'll be better to check pip version in `setup.py`.

PR #458, opened 2016-02-14
#439 #447 Except anydbm instead of gdbm for PY2

### LINKED ISSUE #439: Windows - ImportError: No module named dbm
I have installed thefuck with pip using Python 2.7.11 in Windows. When running `thefuck` from Powershell, I get the error message `ImportError: No module named dbm`. Any ideas?

scorphus:
Fix #439 and close #447

### DIFF
--- a/thefuck/utils.py
+++ b/thefuck/utils.py
@@ -1,4 +1,3 @@
-import dbm
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

PR #462, opened 2016-02-18
#N/A Remove fucked up cmd from history regardless of status

scorphus:
Most fucked up commands are erroneous, but that's not always the case.


nvbn:
Thanks!

PR #472, opened 2016-03-04
Rules: git remote add instead of set-url if remote does not exist

PLNech:
screen shot 2016-03-04 at 13 49 37

### DIFF
--- a/thefuck/rules/git_remote_seturl_add.py
+++ b/thefuck/rules/git_remote_seturl_add.py
@@ -1,0 +1,14 @@
+from thefuck.utils import replace_argument
+from thefuck.specific.git import git_support
+
+
+@git_support
+def match(command):
+    return ('set-url' in command.script and 'fatal: No such remote' in command.stderr)
+
+
+def get_new_command(command):
+    return replace_argument(command.script, 'set-url', 'add')
+
+
+enabled_by_default = True
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ using the matched rule and runs it. Rules enabled by default are as follows:
 * `git_pull_clone` &ndash; clones instead of pulling when the repo does not exist;
 * `git_push` &ndash; adds `--set-upstream origin $branch` to previous failed `git push`;
 * `git_push_pull` &ndash; runs `git pull` when `push` was rejected;
+* `git_remote_seturl_add` &ndash; runs `git remote add` when `git remote set_url` on nonexistant remote;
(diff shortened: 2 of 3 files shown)


mcarton:
Nice, but can we have a test file for that?

PLNech:
@mcarton: I fixed the README test and created a test for this rule.
I wasn't sure which test style I should follow, please tell me if this was not the best way to do so!

scorphus:
Good job, @PLNech :+1: 

Do you mind squashing your three commits into a single one then force-pushing to same branch (your local clone's master, in this case) in order to have a single changeset regarding a single feature?

Thanks!

scorphus:
Also, name the test file after the rule, IOW:
```bash
$ mv test_git_remote_rename_add.py test_git_remote_seturl_add.py
```

thefuck/rules/git_remote_seturl_add.py
```diff
@@ -0,0 +1,11 @@
+from thefuck.utils import replace_argument
+from thefuck.specific.git import git_support
+
+@git_support
```
scorphus: We adopt the style of two lines before a function declaration. Please fix this.

thefuck/rules/git_remote_seturl_add.py
```diff
@@ -0,0 +1,11 @@
...
+def match(command):
+    return('set-url' in command.script and 'fatal: No such remote' in command.stderr)
+
+def get_new_command(command):
```
scorphus: Same here :+1: You might already know but it's always worth mentioning Python's [Code Style](http://docs.python-guide.org/en/latest/writing/style/).

PLNech:
@scorphus: thanks for your comments! Good catch for the test rule name, thanks for the style advice - a squashed commit correcting these points is on its way :)

scorphus:
Awesome! :+1:

PLNech:
Done :)

scorphus:
Superb! :+1: 

Oh! Just in time, @PLNech: I forgot to mention your test lacks a `not_match` test case ([exempli gratia](https://github.com/nvbn/thefuck/blob/5389d0c1068ea5081e0a0cdaf14b45a2c6e0463b/tests/rules/test_git_two_dashes.py#L21)).

PLNech:
@scorphus: Thanks, I updated accordingly and resquashed!

scorphus:
Great. Thank you!

nvbn:
Thanks!

PR #473, opened 2016-03-08
#439 & #447: Remove cache if created with unavailable db

scorphus:
When switching from/to a virtualenv, the database package used to create the cache might be unavailable and an ImportError is raised, such as `ImportError: No module named gdbm`.


scorphus:
Oh! I'm better improving that comment. In a bit.

nvbn:
Thanks!

PR #478, opened 2016-03-15
Add Powershell as shell

MattKotsenas:
- Add Powershell as a supported shell
  - There may be additional functionality to implement, but I've been
    running this way for a month with no known issues
- Update _get_shell to work with Windows
  - _get_shell assumed the parent process would always be the shell
    process, in Powershell the parent process is Python, with the
    grandparent being the shell
  - Switched to walking the process tree so the same code path can be used
    in both places
- Change Ctrl+C on Windows to match Unix

### DIFF
--- a/tests/shells/test_powershell.py
+++ b/tests/shells/test_powershell.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+from thefuck.shells import Powershell
+
+
+@pytest.mark.usefixtures('isfile', 'no_memoize', 'no_cache')
+class TestPowershell(object):
+    @pytest.fixture
+    def shell(self):
+        return Powershell()
+
+    def test_and_(self, shell):
+        assert shell.and_('ls', 'cd') == '(ls) -and (cd)'
+
+    def test_app_alias(self, shell):
+        assert 'function fuck' in shell.app_alias('fuck')
+        assert 'function FUCK' in shell.app_alias('FUCK')
+        assert 'thefuck' in shell.app_alias('fuck')
--- a/thefuck/shells/__init__.py
+++ b/thefuck/shells/__init__.py
@@ -9,12 +9,14 @@ from .fish import Fish
 from .generic import Generic
 from .tcsh import Tcsh
 from .zsh import Zsh
+from .powershell import Powershell
 
 shells = {'bash': Bash,
           'fish': Fish,
           'zsh': Zsh,
           'csh': Tcsh,
-          'tcsh': Tcsh}
+          'tcsh': Tcsh,
+          'powershell': Powershell}
 
 
 def _get_shell():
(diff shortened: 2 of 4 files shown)


scorphus:
Good job, @MattKotsenas. From what I can tell it looks okay, although the two logically distinct changes could be split into two independent commits.

I have a Windows box at home and will give it a shot soon tonight.

Thanks!

MattKotsenas:
I've rebased to split the commit apart. Please let me know how it goes!

thefuck/shells/powershell.py
```diff
@@ -0,0 +1,17 @@
+from .generic import Generic
+
+class Powershell(Generic):
```
scorphus: It's a good style we follow to have 2 lines before a class definition.

thefuck/shells/powershell.py
```diff
@@ -0,0 +1,17 @@
...
+    def app_alias(self, fuck):
+        return  'function ' + fuck + ' { \n' \
```
scorphus: Also, there are 2 spaces between `return` and `'`. Please remove that one and align it linedown.

tests/shells/test_powershell.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+
+import os
```
scorphus: This import isn't used. Can you please remove it?

thefuck/shells/__init__.py
```diff
@@ -9,20 +9,37 @@
...
+
+        if name in shells:
+            return shells[name]()
+
+        try:
+            proc = proc.parent()
+        except TypeError:
+            proc = proc.parent
+
+    return Generic()
```
scorphus: Here you're creating an instance of `Generic`. `_get_shell` should return the class itself.

scorphus:
@MattKotsenas You did great with the commit split! Thanks! Can you please fix that minor issues? I'm glad explaining any of them.

Also, this is how my `$profile` reads:
```powershell
# function fuck {
#     $fuck = $(thefuck (get-history -count 1).commandline)
#     if ("$fuck".startswith("echo")) {
#         "$fuck".substring(5)
#     } else {
#         if ($fuck) {
#             iex "$fuck"
#         }
#     }
# }

iex "thefuck --alias"
```
But this is what I got:
```
(py27) PS C:\Users\IEUser> fuck
Seems like fuck alias isn't configured!
Please put iex "thefuck --alias" in your $profile.
More details - https://github.com/nvbn/thefuck#manual-installation
(py27) PS C:\Users\IEUser>
```

What am I missing? I'm on a Windows 8.1 by the way.

Thanks.

MattKotsenas:
Thanks @scorphus for the feedback. Stylistic feedback addressed and squashed into previous commit.

For the alias I'm not sure what's going on. If I do
```
iex "thefuck --alias" >> $profile
```

And restart PowerShell (necessary to pick up changes) I get the expected behavior. Am I interpreting the prompt wrong? Thanks!

nvbn:
Thanks!

PR #483, opened 2016-03-21
Update install procedure for pip

shakaran:
Avoid the warning:

    The directory '/home/someuser/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.


nvbn:
Thanks!

PR #492, opened 2016-04-07
Treat overridden aliases in a better way

scorphus:
As soon as we agree on this, I'll update the `README.md` file and add a Wiki entry regarding overridden aliases in Fish.


nvbn:
Thanks!

PR #495, opened 2016-04-21
#493: Add `put_to_history` back to Fish

### LINKED ISSUE #493: Correct command not being pushed to fish history
I'm not entirely sure what's going on, but if I run `puthon`, invoke `fuck` to get `python`, my history output is
```
fuck ##the fuck that I typed to replace puthon with python
foo #previous commands
bar
```
but nothing of the fixed command.

scorphus:
Fish Shell does not support appending to history yet

Fix #493

### DIFF
--- a/thefuck/types.py
+++ b/thefuck/types.py
@@ -280,6 +280,8 @@ class CorrectedCommand(object):
         """
         if self.side_effect:
             compatibility_call(self.side_effect, old_cmd, self.script)
+        if settings.alter_history:
+            shell.put_to_history(self.script)
         # This depends on correct setting of PYTHONIOENCODING by the alias:
         logs.debug(u'PYTHONIOENCODING: {}'.format(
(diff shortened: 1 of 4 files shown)


scorphus:
@nvbn: If there's something that could be done better/another way, just let me know.

thefuck/types.py
```diff
@@ -280,6 +280,8 @@ def run(self, old_cmd):
         """
         if self.side_effect:
             compatibility_call(self.side_effect, old_cmd, self.script)
+        if settings.alter_history:
```
scorphus: @nvbn: I thought about adding another `check if Fish` here but then thought this could be decided by the Shell classes implementation. Maybe a comment on `Generic.put_to_history()` would be a good idea, for future reference. What do you think?

nvbn: @scorphus I've added comment.

nvbn:
Thanks!

Merged in d6c2c72. Strange that github didn't see that =)

PR #515, opened 2016-06-11
Add a new rule `git_rebase_no_changes`.

TheJakeSchmidt:
This rule fixes the case where you've run "git rebase --continue", as instructed by "git status", only to be rejected because there are no changes in the commit. The fix is rote; you run "git rebase --skip" instead of "git rebase --continue".


nvbn:
Thanks!
I've made a few changes.

PR #521, opened 2016-06-25
ui: accept 'q' as quit character

b1101:
'q' is a standard character used in traditional UNIX environment used for 'quit', so it makes sense to support it, in my opinion

Added tests and ammended the previous commit, added missing tests for 'j' and 'k'


nvbn:
Thanks!

PR #523, opened 2016-06-27
Add semi-colons in powershell alias

MattKotsenas:
Add semi-colons in powershell alias so that if line breaks get lost the function can still be invoked directly. This makes it possible to add thefuck to the current session by running

```powershell
iex "$(thefuck --alias)"
```

which mirrors the eval syntax in bash


MattKotsenas:
We should also update the wiki once this update gets pushed to ```pip```. I don't want to do that ahead-of-time because it would break new users.

nvbn:
Thanks!

PR #525, opened 2016-06-29
Avoid importing pip

tdsmith:
Pip doesn't have a stable Python API and shouldn't be `import`ed, and may not be used to install this package at all.


nvbn:
Thanks!

Issue #526, opened 2016-06-29
`brew udpate` corrects to `brew outdated`, not `brew update`


josephfrazier:
Information requested in CONTRIBUTING.md:

* the output of thefuck --version
```
thefuck --version
The Fuck 3.10 using Python 2.7.5
```

* your shell and its version (bash, zsh, Windows PowerShell, etc.);
```
$SHELL --version
GNU bash, version 4.3.46(1)-release (x86_64-apple-darwin13.4.0)
```

* how to reproduce the bug:
```
brew udpate
fuck
```

* the output of The Fuck with THEFUCK_DEBUG=true exported (typically execute export THEFUCK_DEBUG=true in your shell before The Fuck):

https://gist.github.com/josephfrazier/5fdf8d058f7c91a34178471c87e2d990

* if the bug only appears with a specific application, the output of that application and its version;
```
brew --version
Homebrew 0.9.9 (git revision db76a; last commit 2016-06-29)
Homebrew/homebrew-core (git revision 14be2; last commit 2016-06-29)
```

PR #527, opened 2016-06-29
Add `brew update` to Homebrew commands

josephfrazier:
`brew update` is implemented in shell instead of ruby, so
`_get_brew_commands` needs to list .sh files as well as .rb

Resolves https://github.com/nvbn/thefuck/issues/526


nvbn:
Thanks!

PR #529, opened 2016-07-07
Added "contains work" error for git push

yeahbert:
I'm not 100% sure what the difference between "because the tip of your current branch is behind" and "because the remote contains work that you do not have locally" is, but since git pull seems to be a valid solution for both, I added it to the rule.


nvbn:
Thanks!

PR #532, opened 2016-07-25
start work on -y

### LINKED ISSUE #531: allow for a flag that executes immediately
I often do the same `git push` to a new brach that needs `--set-upstream origin ...` to create a new branch. `fuck` can fix this, but it forces me to press `enter` even though I know what input it needs. I _could_ get an alias that creates a branch immediately and don't require `fuck` at all anymore, but I'd prefer to be able to `fuck -y` that executes the suggested command immediately.

Haroenv:
Fixes #531 

[WIP]

@scorphus:  I have never programmed in Python before, is this where I add the option, or is it somewhere else?

TODO
- [ ] add flag
- [ ] make flag set `settings.require_confirmation: False`
- [ ] make flag execute the rest of the command
- [ ] add tests

### DIFF
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -72,6 +72,11 @@ def main():
     parser.add_argument('-a', '--alias',
                         action='store_true',
                         help='[custom-alias-name] prints alias for current shell')
+    parser.add_argument('-y',
+                        # somehow put settings.require_confirmation: False
+                        # Then execute command
+                        action='command',
+                        help='executes the fixed command without confirmation')
     parser.add_argument('command',
                         nargs='*',
                         help='command that should be fixed')
--- a/thefuck/main.py
+++ b/thefuck/main.py
@@ -72,11 +72,6 @@ def main():
     parser.add_argument('-a', '--alias',
                         action='store_true',
                         help='[custom-alias-name] prints alias for current shell')
-    parser.add_argument('-y',
-                        # somehow put settings.require_confirmation: False
-                        # Then execute command
-                        action='command',
-                        help='executes the fixed command without confirmation')
     parser.add_argument('command',
                         nargs='*',
                         help='command that should be fixed')
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -26,6 +26,9 @@ class Fish(Generic):
             alter_history = ''
         # It is VERY important to have the variables declared WITHIN the alias
         return ('function {0} -d "Correct your previous console command"\n'
+                '  if contains -- -y $argv\n'
+                '    set THEFUCK_REQUIRE_CONFIRMATION 0\n'
+                '  end\n'
                 '  set -l fucked_up_command $history[1]\n'
                 '  env TF_ALIAS={0} PYTHONIOENCODING=utf-8'
                 ' thefuck $fucked_up_command | read -l unfucked_command\n'
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -27,6 +27,7 @@ class Fish(Generic):
         # It is VERY important to have the variables declared WITHIN the alias
         return ('function {0} -d "Correct your previous console command"\n'
                 '  if contains -- -y $argv\n'
+                '    set previous_confirmation_value THEFUCK_REQUIRE_CONFIRMATION\n'
                 '    set THEFUCK_REQUIRE_CONFIRMATION 0\n'
                 '  end\n'
                 '  set -l fucked_up_command $history[1]\n'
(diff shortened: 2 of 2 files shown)


scorphus:
To have it implemented the way we discussed earlier – at shell alias/function level – it would involve a rewriting the shell alias – i.e. [`Fish.app_alias()`](https://github.com/nvbn/thefuck/blob/51415a5cb1ca6955fb99908e7d0e7bf012a66312/thefuck/shells/fish.py#L21) – in order to make it conditionally set `THEFUCK_REQUIRE_CONFIRMATION` to `0` prior to calling `thefuck ...`.

Haroenv:
Aha, it's there, I interpreted your comment differently!

scorphus:
Great!

ps.: I'm pondering how this could be done in the Bash alias, for instance.

Haroenv:
Okay, so @scorphus, I found how to add a fish argument, but where should this be documented, tested?

Haroenv:
For powershell, something like [switch](https://social.technet.microsoft.com/Forums/windowsserver/en-US/2961ae65-d7d9-4c1d-bdf2-505273925ccc/advanced-functions-flags?forum=winserverpowershell) could be used (but I don't know how it finds out what the name of the flag is

Haroenv:
In `bash` it'll need to be replaced by a [function](http://apple.stackexchange.com/a/51010/98431)

scorphus:
> […] I found how to add a fish argument, but where should this be documented, tested? […]

There are some [funcitonal tests ](https://github.com/nvbn/thefuck/tree/51415a5cb1ca6955fb99908e7d0e7bf012a66312/tests/functional) that could include tests for the new `-y` functionality.

> For powershell, something like [switch](https://social.technet.microsoft.com/Forums/windowsserver/en-US/2961ae65-d7d9-4c1d-bdf2-505273925ccc/advanced-functions-flags?forum=winserverpowershell) could be used (but I don't know how it finds out what the name of the flag is

Looks like it'll work. Maybe @MattKotsenas could chime in and shed some some light on the matter?

> In `bash` (probably the other shells as well, seeing they're aliases instead of functions too) it'll need to be replaced by a [function](http://apple.stackexchange.com/a/51010/98431)

Yeah, that's the critical change, I think. Not sure how much it impacts on the way TheFuck is set up by users and [tools](https://github.com/Bash-it/bash-it/blob/7415134878b8b01015c9a9fdbc66b784db02ff65/aliases/available/fuck.aliases.bash) alike.

Perhaps there should be a `--function` argument that always creates a `function`. Then we can leave `--alias` functionality untouched. What do you think?

Haroenv:
I just realised that this would unset the confirmation for everything after once `-y`, will need `else` and setting it back to the default. (or setting it to the default at the end of the command maybe?)

scorphus:
Here's [another example](https://github.com/robbyrussell/oh-my-zsh/blob/a7e30b26baa94bac99d9d05cf642bd1942ae1787/plugins/thefuck/thefuck.plugin.zsh#L7) of a tool that uses TheFuck.

Thinking it again, on most cases there won't be any problems creating a `function` instead of an `alias`. Let me recall how users used to set them.

scorphus:
> […] setting it back to the default […]

That might be unnecessary. It should be an _in-command-variable_, something like:

```fish
# middle of Fish alias
if dash_y
  env THEFUCK_REQUIRE_CONFIRMATION=0 TF_ALIAS=fuck PYTHONIOENCODING=utf-8 thefuck <cmd>
else
  env TF_ALIAS=fuck PYTHONIOENCODING=utf-8 thefuck <cmd>
end
# …
```

scorphus:
Since [aliases changed in 1.34](https://github.com/nvbn/thefuck#update) and are defined with `eval "$(thefuck --alias)"` I think we might be okay changing it to a `function`.

/cc @nvbn any concerns?

scorphus:
Oh-My-Fish [TheFuck plugin](https://github.com/oh-my-fish/plugin-thefuck/blob/86f7e1b720f8395964f348f601b4b8fd9a1cf671/functions/fuck.fish), for example, will need to be updated to support this feature, although it would continue to work nonetheless.

scorphus:
@Haroenv: the condition can be written like this:

```fish
contains -- "-y" $argv; and set -lx THEFUCK_REQUIRE_CONFIRMATION 1
```

* `l`: sets the variable locally – it is bound to the funciton scope only
* `x`: exports the variable to sibling processes – `thefuck` will be able to read it

nvbn:
Maybe it can be a bit simpler, just a different alias like:
```bash
alias fuckit='THEFUCK_REQUIRE_CONFIRMATION=0 fuck'
```
?

Haroenv:
That seems like a great idea, should need to check if that saves the variable or if it uses it just once though. Using fuckit as a new alias makes sense

Issue #534, opened 2016-08-10
open with mac on a file the does not exist does not just create the damn file


mlk:
MacBook-Pro:~ mlk$ open idontexist
The file /Users/mlk/idontexist does not exist.
MacBook-Pro:~ mlk$ fuck
No fucks given
MacBook-Pro:~ mlk$ 

Correct response should be
touch idontexist && open idontexist

mlk:
Also maybe fuck after a (seemingly successful) should offer 
open -e file

scorphus:
> Correct response should be
> touch idontexist && open idontexist

Good one. `mkdir idontexist && open idontexist` could be a valid correction too. I can code it 😃 

> Also maybe fuck after a (seemingly successful) should offer 
> open -e file

Hmm... TheFuck is intended to *correct* your last command. It *expects* it to fail.

mlk:
> Hmm... TheFuck is intended to correct your last command. It expects it to fail.

The problem with open is that it will attempt to open it with the thing that is registered to open it. Which may fail (:shakefist: @ tunnelblick and its opening of .conf), however I think at that point open has exited. I've not looked at the source code for a while, does TheFuck first check for a none-zero exit command?

scorphus:
> […] does TheFuck first check for a none-zero exit command?

Not exactly. It runs the command regardless of the exit status. Its interest is in the command's output ( both `STDERR` and `STDOUT`).

So, imagine you hit `open index.html`. `open` does not fail here. Instead, it successfully opens `index.html` with the default web browser. Assuming TheFuck could fix that and turn it into `open -e index.html`, you then think *“Fuck!”*, close the browser and hit `fuck`. TheFuck then runs your previous command – `open` – to get its output, which in case loads up `index.html` with the default web browser yet again and exits right away, flawlessly, with no output. With that info, TheFuck prompts you with the *fixed* command, but you're left with `index.html` opened in the default web browser.

Perhaps some rules could try and match the command script alone, regardlessly of the output, and TheFuck could run a first round of those rules before running the command itself. It could also base on the last command's exit status. This way we would assume the user knows what he's trying to unfuck. Not really sure how well this could work, though, I haven't thought much about it.

scorphus:
I'm gonna turn this last paragraph in a POC, when time allows.

mlk:
Oh yeah it re-runs the command to get the output.

nvbn:
Fixed in previous releases.

PR #535, opened 2016-08-12
#534: Improve open rule

scorphus:
This change improves the `open` rule and its tests and implement one of the features suggested by @mlk on #534.


nvbn:
Thanks!

PR #540, opened 2016-08-21
create new config dirs in $HOME instead of $CWD

kthrift:
fix: new config dirs created in $HOME/.config/thefuck instead of $CWD/.config/thefuck
fix: use correct path in warning message when XDG_CONFIG_HOME defined

### DIFF
--- a/thefuck/conf.py
+++ b/thefuck/conf.py
@@ -40,16 +40,21 @@ class Settings(dict):
                     settings_file.write(u'# {} = {}\n'.format(*setting))
 
     def _get_user_dir_path(self):
-        # for backward compatibility, use `~/.thefuck` if it exists
-        legacy_user_dir = Path('~/.thefuck').expanduser()
+        """returns Path object representing the user config resource"""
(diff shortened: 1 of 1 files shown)


kthrift:
This pull request fixes a bug in the _get_user_dir_path method that created a
directory named '~' in the current working directory unless either of
the following were true:

1. the legacy user config directory, named $HOME/.thefuck, existed
2. the environment variable XDG_CONFIG_HOME existed and contained /fully expanded paths (i.e. $HOME/.config instead of ~/.config)

The simplest change to fix the issue would be to change line 52 of
thefuck/conf.py as so:
    `--- return Path(xdg_config_dir).joinpath('thefuck')`
    `+++ return Path(xdg_config_dir).expanduser().joinpath('thefuck')`

However, this changeset includes some minor refactoring to provide:
- a single 'return value' statement
- fixes the logging statement for the case where XDG_CONFIG_HOME is set to something other than $HOME/.config

nvbn:
Thanks!

> a single 'return value' statement

In short functions I prefer to use multiple `return` instead of changing variable.

PR #548, opened 2016-09-14
add aliases rule that resolves aliases

awesomefireduck:
This aims to add a rule that resolves shell aliases and replaces them with the full version and then runs thefuck on that with the already supplied outputs.
(i.e. replaces the alias in command.script with the full version.)

I am not sure how to test this further. Could you please give me  a hand?
I am currently not able to call the `get_corrected_command` function

TODO:
- import the needed function(s) correctly
- not hardcode the bash shell: use $SHELL or run command in the already open shell(?)
- write test (as of writing a lot of the tests seem to fail, is this correct?)

### DIFF
--- a/thefuck/rules/aliases.py
+++ b/thefuck/rules/aliases.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python2
+import subprocess
+import re
+from thefuck.corrector import get_corrected_command
+
+
+enabled_by_default = True
+priority = 100000
+
+
+def list_aliases():
+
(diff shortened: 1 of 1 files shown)


nvbn:
thefuck already expands aliases, for example:

```bash
➜ gp
fatal: The current branch test has no upstream branch.

    git push --set-upstream origin test

➜ fuck
git push --set-upstream origin test [enter/↑/↓/ctrl+c]
```

awesomefireduck:
Perhaps I just need to update `thefuck`...

PR #552, opened 2016-09-29
Custom rule: Remove trailing cedillas

wikiti:
### Summary

Many programmers use european keyboards, like this one:

BÉPO Keyboard Layout

Some users (like me), have some fat-brute-fingers, and, when pressing the `enter` key, we also press that tedious `ç` key. I've created this new rule to fix those typos.

### Features
- [x] Create a new rule to match those cases, and fix them.
- [x] Create a new test.

### Questions

- Should the priority be changed to a lower value?

### References
- [Cedilla in Wikipedia](https://en.wikipedia.org/wiki/Cedilla)
- [Keyboard layouts in Wikipedia](https://en.wikipedia.org/wiki/Keyboard_layout)


slavaGanzin:
You should have a massive pinky...

wikiti:
@slavaGanzin This:

img_20160929_111535

nvbn:
Thanks!

Issue #553, opened 2016-09-29
thefuck does not suggest `brew uninstall --force` when relevant


josephfrazier:
When I `brew uninstall` a package with multiple versions, I usually want to uninstall all of them. However, `thefuck` does not offer that as a suggestion. For example:

```shell
[josephfrazier@Josephs-MBP ~] $
brew uninstall tbb
Uninstalling /usr/local/Cellar/tbb/4.4-20160916... (118 files, 1.9M)
tbb 4.4-20160526, 4.4-20160722 are still installed.
Remove all versions with `brew uninstall --force tbb`.
[josephfrazier@Josephs-MBP ~] $
fuck
brew info tbb [enter/↑/↓/ctrl+c]
```

Instead, `thefuck` should suggest `brew uninstall --force tbb`. I'll submit a pull request fixing this, but I wanted to make sure it has an issue first.

PR #555, opened 2016-09-29
Suggest `brew uninstall --force` when relevant

josephfrazier:
Resolves https://github.com/nvbn/thefuck/issues/553


nvbn:
Thanks!

PR #556, opened 2016-09-30
Suggest `brew link --overwrite --dry-run` when relevant

josephfrazier:
This makes it easier to see which files would be overwritten by
`brew link --overwrite`


nvbn:
Thanks!

PR #557, opened 2016-09-30
Add sudo rule for Aura

### LINKED ISSUE #543: Adding sudo works for `aura -Sy` but not `aura -Ay`
`fuck` is unable to add `sudo` to an `aura -Ay` command:
```
$ aura -Ay foobar-beta-git  # from AUR
aura >>= You have to use `sudo` for that.
$ fuck
No fucks given
```

But works as expected for `aura -Sy`:
```
$ aura -Sy foobar  # pacman alias
error: you cannot perform this operation unless you are root.
aura >>= Please check your input.
$ fuck
sudo aura -Sy foobar [enter/↑/↓/ctrl+c]
```
It's slightly annoying anyway that the `aura` outut is different in these cases, but is it possible for `thefuck` to work-around? Or is the only way for `aura` to give a stderr message containing "root"?

OJFord:
When installing from Arch User Repository without root:
    aura >>= You have to use `sudo` for that.

This PR adds the slightly more general, but unambiguous, "use `sudo`".

It might be helpful to also add "use sudo" (no backticks) - but I haven't since I don't know for sure it's in use.

Closes #543.

### DIFF
--- a/thefuck/rules/sudo.py
+++ b/thefuck/rules/sudo.py
@@ -19,7 +19,8 @@ patterns = ['permission denied',
             'you don\'t have access to the history db.',
             'authentication is required',
             'edspermissionerror',
-            'you don\'t have write permissions']
+            'you don\'t have write permissions',
+            'use `sudo`']
 
 
 def match(command):


nvbn:
Thanks!

PR #560, opened 2016-10-03
Suggest `git diff --no-index` when relevant

josephfrazier:
This makes it easier to use `git diff` on untracked files.


nvbn:
Thanks!

PR #561, opened 2016-10-03
Suggest `ag -Q` when relevant

josephfrazier:
This detects when `ag` suggests the `-Q` option, and adds it.

### DIFF
--- a/thefuck/rules/ag_literal.py
+++ b/thefuck/rules/ag_literal.py
@@ -1,0 +1,11 @@
+from thefuck.utils import for_app
+from thefuck.utils import replace_argument
+
+
+@for_app('ag')
+def match(command):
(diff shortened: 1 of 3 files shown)


thefuck/rules/ag_literal.py
```diff
@@ -0,0 +1,11 @@
+from thefuck.utils import for_app
+from thefuck.utils import replace_argument
+
+
+@for_app('ag')
+def match(command):
+    return 'run ag with -Q' in command.stderr
```
scorphus: Isn't `endswith()` a slightly better option than `in`?

josephfrazier:
Thanks for the review, @scorphus! I've made some changes addressing your comments.

nvbn:
Thanks!

Issue #546, opened 2016-09-06
suggestion for "man command" rule: "command --help"


waldyrious:
I don't think thefuck should attempt to change the man section if no section was provided in the first place -- at least, I don't think that's what most people would expect. It's probably more useful to rename that rule "man_section", and use the "man" rule to `run command --help` instead.

PR #562, opened 2016-10-03
Suggest `foo --help` when `man foo` shows no pages

josephfrazier:
This is along the lines of what @waldyrious suggested in
https://github.com/nvbn/thefuck/issues/546, but it just adds a new
suggestion rather than replacing the other ones.

### DIFF
--- a/tests/rules/test_man.py
+++ b/tests/rules/test_man.py
@@ -23,7 +23,7 @@ def test_not_match(command):
 
 
 @pytest.mark.parametrize('command, new_command', [
-    (Command('man read'), ['man 3 read', 'man 2 read', 'read --help']),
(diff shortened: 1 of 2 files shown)


waldyrious:
@josephfrazier thanks! What would this work like? Would one have to type "fuck" 3 times to get to the `--help` behavior?

josephfrazier:
When multiple suggestions are available, you can scroll between them using the up/down arrow keys. In this case, you'd type 'fuck', then press the up arrow key until you see the `--help` option.

However, after further experimentation, I learned that `man` searches all sections if you don't explicitly provide one in your command, so I'll probably update this pull request so that it doesn't suggest a section when there was none provided, as you had originally requested in #546.

nvbn:
Thanks!

PR #563, opened 2016-10-05
Run flake8 in Travis, fix some errors

josephfrazier:


josephfrazier:
Cheers @scorphus! It was your comments on #561 that inspired me to do this. I figured it might save some effort going forward.

tests/rules/test_fix_file.py
```diff
@@ -227,9 +227,9 @@ def test_get_new_command(mocker, monkeypatch, test):
     mocker.patch('os.path.isfile', return_value=True)
     monkeypatch.setenv('EDITOR', 'dummy_editor')
 
-    cmd = Command(script=test[0], stdout=test[4], stderr=test[5])
-    #assert (get_new_command(cmd, Settings({})) ==
-    #    'dummy_editor {} +{} && {}'.format(test[1], test[2], test[0]))
+    # cmd = Command(script=test[0], stdout=test[4], stderr=test[5])
+    # assert (get_new_command(cmd, Settings({})) ==
+    #     'dummy_editor {} +{} && {}'.format(test[1], test[2], test[0]))
```
scorphus: I think we can remove this test entirely. If we ever need it, Git will give it back to us like magic.

thefuck/main.py
```diff
@@ -1,8 +1,6 @@
 # Initialize output before importing any module, that can use colorama.
 from .system import init_output
 
-init_output()
```
scorphus: This one is necessary. Looks like we'll need `# noqa: E402` in every of the following import lines.

tests/rules/test_ln_s_order.py
```diff
@@ -38,4 +28,5 @@ def test_not_match(file_exists, script, stderr, exists):
     ('ln dest source -s', 'ln source -s dest')])
 def test_match(script, result):
     stderr = get_stderr('source')
+    assert match(Command(script, stderr=stderr))
     assert get_new_command(Command(script, stderr=stderr)) == result
```
scorphus: Nice one.

tests/rules/test_git_push_pull.py
```diff
@@ -68,12 +63,7 @@ def test_not_match(command):
...
-     'git pull nvbn master && git push nvbn master')])
-def test_get_new_command(command, output):
-    assert get_new_command(command) == output
-
-
-@pytest.mark.parametrize('command, output', [
+     'git pull nvbn master && git push nvbn master'),
     (Command(script='git push', stderr=git_err2), 'git pull && git push'),
     (Command(script='git push nvbn', stderr=git_err2),
      'git pull nvbn && git push nvbn'),
```
scorphus: Good catch.

josephfrazier:
Thanks again for the comments, @scorphus! I made some more changes addressing them :)

scorphus:
Awesome, @josephfrazier!

Well, I feel I can push this even further 😃  As you're in the process of fixing these issues and changes haven't yet made into master, what would you say if I suggested you to remove a couple of commits and squash those which are logically similar/related?

Commit 85a3e34 is a revert of 4d9f177, both could be removed. Also, all EXYZ-related commits could be joined together, leaving one single commit per violation – E123, E225, E231, E265, E302, E402, E711 and E731.

What do you think about getting a cleaner history?

josephfrazier:
Good call @scorphus, I did a lot more error-fixing and cleaned up the history :D

tests/rules/test_fix_file.py
```diff
@@ -14,47 +14,47 @@
...
-"""),
+"""),  # noqa
 
 ('clang a.c', 'a.c', 3, 1, '',
 """
 a.c:3:1: error: expected expression
 }
 ^
-"""),
+"""),  # noqa
```
scorphus: How about this?

```python
tests = (
    ('gcc a.c', 'a.c', 3, 1, '',
     """
a.c: In function 'main':
a.c:3:1: error: expected expression before '}' token
 }
  ^
"""),

    ('clang a.c', 'a.c', 3, 1, '',
     """
a.c:3:1: error: expected expression
}
^
"""),
```

I, personally, prefer this over `# noqa` in every single line. What do you think?

josephfrazier: I considered that alternative as well. However, after noticing that some multiline strings also have trailing whitespace (see 432878bd773319cf034d3070697e8601d4201b6d), I thought it would be easier to just use `# noqa` to ignore both the whitespace errors and the indent errors. My thought process was that it wouldn't be good to edit the string to remove the trailing whitespace, since it was likely pasted from a terminal and we should preserve its formatting. As far as I know, there's no way to have flake8 ignore _just_ the trailing whitespace in multiline strings while still emitting indent errors for them, unfortunately.

Of course, preserving trailing whitespace inside multiline strings may not be all that important (it's hard to imagine a rule that relies on trailing whitespace for matching to work correctly), so maybe we should remove it, remove the `# noqa` comments, and indent the leading triple quotes as you're suggesting. Here's a commit showing what that might look like: https://github.com/josephfrazier/thefuck/commit/25efd1b670c8f9694f6e5617fe4c38666f9b0704. I'm not sure if I like the way [tests/rules/test_fix_file.py](https://github.com/josephfrazier/thefuck/blob/25efd1b670c8f9694f6e5617fe4c38666f9b0704/tests/rules/test_fix_file.py) looks, though :/

scorphus: > I thought it would be easier to just use # noqa to ignore both the whitespace errors and the indent errors.

Yeah. At least until flake8 doesn't ignore just what it's told to. Maybe we're – or I am – getting a bit too meticulous here. These are test files, after all. Anyway, I guess that just appending `# noqa` to the one line where the `tests = (...)` tuple definition ends, is sufficient.

josephfrazier: Oh, so it does! I think that's the best option so far, and I've updated the branch accordingly :D

josephfrazier:
@nvbn, do you have any thoughts on this? Is there anything else that needs to be done to land it (other than fixing files that have been changed on the `master` branch)?

nvbn:
@josephfrazier, ooops, sorry, I just forgot about this pr.

I'll check it in the nearest time and will merge it.

Thansk.

josephfrazier:
Great, thanks! I just pushed some changes that fix the files that had been updated on the master branch, but haven't rebased them yet. If everything looks good to you, let me know and I can `git rebase -i --autosquash master` to combine the `fixup!` commits with their referents and get rid of that `Merge branch 'master' into flake8` commit.

nvbn:
Thanks! Merged in master.

josephfrazier:
Cheers, hopefully this will make it easier to ensure future contributions adhere to the code style.

PR #564, opened 2016-10-06
Use official Python images for Docker tests

josephfrazier:
This reduces build times by about 2 minutes, since Travis no longer has to install python/pip/git/setuptools on top of the Ubuntu images:

Before (7 min 1 sec): https://travis-ci.org/nvbn/thefuck/jobs/164603738
After (5 min 15 sec): https://travis-ci.org/josephfrazier/thefuck/jobs/165560110


nvbn:
Thanks!

PR #571, opened 2016-10-31
bash: always honor alter_history setting

josephfrazier:
This ensures that even if the command suggested and run by `thefuck`
fails, it will still be added to the history, allowing the user to tweak
it further (or run `fuck` again) if desired.

Note that the fish shell appears to already behave this way.


andrewcates:
Unsubscribe

nvbn:
Thanks!

PR #572, opened 2016-10-31
Suggest `ls -A` when `ls` has no output

josephfrazier:
For example:

    $ ls dotfiles/
    $ fuck
    ls -A dotfiles/ [enter/↑/↓/ctrl+c]


nvbn:
Thanks!

PR #580, opened 2016-11-27
bash: fix parsing of command substitution

josephfrazier:
This addresses bad corrections like the following (note the position of the `-p` flag):

```bash
thefuck 'git log $(git ls-files thefuck | grep python_command) -p'
git log $(git ls-files thefuck | grep -p python_command) [enter/↑/↓/ctrl+c]
```

After the changes in this PR, the above command looks like this instead:

```bash
thefuck 'git log $(git ls-files thefuck | grep python_command) -p'
git log -p $(git ls-files thefuck | grep python_command) [enter/↑/↓/ctrl+c]
```

This is done by using [`bashlex`](https://github.com/idank/bashlex) to parse bash, rather than `shlex`. `bashlex` does have a few [limitations](https://github.com/idank/bashlex#limitations), so `shlex` is still used as a fallback in case `bashlex` fails.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ elif (3, 0) < version < (3, 3):
 
 VERSION = '3.11'
 
-install_requires = ['psutil', 'colorama', 'six', 'decorator']
+install_requires = ['psutil', 'colorama', 'six', 'decorator', 'bashlex']
 extras_require = {':python_version<"3.4"': ['pathlib2'],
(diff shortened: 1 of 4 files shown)


josephfrazier:
Hmm, it looks like AppVeyor had some brief bugginess earlier. Here's a successful rebuild of the same commit: https://ci.appveyor.com/project/josephfrazier/thefuck/build/1.0.14

nvbn:
Thanks!

PR #596, opened 2017-01-28
Add git_tag_force rule

josephfrazier:
This adds `--force` to `git tag` when needed. For example:

    $ git tag alert
    fatal: tag 'alert' already exists
    $ fuck
    git tag --force alert [enter/↑/↓/ctrl+c]
    Updated tag 'alert' (was dec6956)
    $


nvbn:
Thanks!

PR #599, opened 2017-01-28
Added support for "not installed" message in apt_get

juzim:
The error message "The program 'vim' is currently not installed. You can install it by typing: sudo apt install vim" has the same solution as "vim: command not found", so I added it as a case for installing the program with apt.


nvbn:
Thanks!

PR #604, opened 2017-02-08
add support for colemak style vi bindings

### LINKED ISSUE #603: vi-style key bindings for colemak
this tools allows j/k (vi style) arrow keys to be used for command selection.

this is all nice, but this does not work with the colemak keyboard layout.

what the fuck.

the n/e keys should also be accepted.

wbolster:
this allows e/n in addition to j/k (same places on the keyboard on colemak and qwerty) to be used as arrow keys when selecting a command from the suggested fixups.

fixes #603

### DIFF
--- a/thefuck/ui.py
+++ b/thefuck/ui.py
@@ -12,9 +12,10 @@ def read_actions():
     while True:
         key = get_key()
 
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

Issue #371, opened 2015-09-10
Fuck has a ~1s lag now


MightyPork:
It worked really good before, but now when I do something like this, I get lag:

```
$ git brnahc
git: 'brnahc' is not a git command. See 'git --help'.

Did you mean this?
        branch
$ fuck
--- 1 second lag ---
git branch [enter/↑/↓/ctrl+c]
* dev
  master
  testing
```

I have version 3.0.

Clues what to try to make it work fast again? Not really sure where is the problem.

Here's also my $PATH if it matters

```
/home/ondra/.gem/ruby/2.2.0/bin:/home/ondra/GNUstep/Tools:/usr/local/sbin:/usr/local/bin:/usr/bin:/opt/android-sdk/platform-tools:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/ondra/bin:/home/ondra/.composer/vendor/bin
```

I'm on Arch Linux (64bit) with python 3.4.3

nvbn:
Can you remove `~/.thefuck/.thefuck-cache` and measure time again?

MightyPork:
Small improvement, I tried "time fuck" and pressed ^C just when it appeared, and got 0.6s. So I think something like 0.5s. Maybe that's normal, not sure - I have SSD though, should be really fast...

nvbn:
And can you post output of:
```bash
export THEFUCK_DEBUG=true
git br
fuck
```
?

MightyPork:
Here it is:

```
$ git br
git: 'br' is not a git command. See 'git --help'.

Did you mean one of these?
	branch
	var
$ fuck
DEBUG: Run with settings: {'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'no_colors': False,
 'priority': {},
 'require_confirmation': False,
 'rules': [],
 'wait_command': 3}
DEBUG: Received stdout: 
DEBUG: Received stderr: 14:49:49.147308 git.c:558               trace: exec: 'git-br'
14:49:49.147380 run-command.c:347       trace: run_command: 'git-br'
git: 'br' is not a git command. See 'git --help'.

Did you mean one of these?
	branch
	var

[...]

Issue #353, opened 2015-09-01
Command is slow


ukch:
Running the 'fuck' command takes a good 3-4 seconds on my machine, making it very slow. I am running the default version from Pip (2.8) on Python 2.7.10 on Debian Sid. If it matters, my shell is zsh.

mcarton:
For me most of the time is actually spent in my shell to get the alias list:
```
% time zsh -ic alias > /dev/null
zsh -ic alias > /dev/null  0.32s user 0.05s system 96% cpu 0.384 total
% time thefuck test               
No fucks given
thefuck test  0.54s user 0.09s system 97% cpu 0.640 total
```
can you execute these command for comparison please? And also the output of
```
export THEFUCK_DEBUG=true
thefuck some-fucked-cmd
```

---
As for the alias list, if that's really a problem and not just my `.zshrc` which need some cleaning up maybe we could make alias resolution optional (since we cannot reasonably ask every user to clean up their `.whatevershellrc` :p).

hcgraf:
I had the same problem, in my case the history rule was very slow.
Try to use the version for python 3, the history rule seems to be MUCH faster…

pip uninstall thefuck
pip3 install thefuck

nvbn:
@mcarton I've added cache for aliases, now it's slow only in the first run in a session..

nvbn:
@ukch try to upgrade to 2.9, it's faster.

ukch:
I haven't tried either upgrading or using Python 3 (I will try that next) but here's the output of the commands that @mcarton posted:

    $ time zsh -ic alias > /dev/null
    zsh -ic alias > /dev/null  0.45s user 0.08s system 88% cpu 0.605 total
    $ time thefuck test
    ls test [enter/ctrl+c]
    ls test
    thefuck test  3.31s user 0.10s system 88% cpu 3.859 total
    $ env THEFUCK_DEBUG=true thefuck apt-get install foo
    DEBUG: Run with settings: {'debug': True,
     'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
     'no_colors': False,
     'priority': {},
     'require_confirmation': True,
     'rules': [],
     'wait_command': 3}
    DEBUG: Received stdout: 
    DEBUG: Received stderr: E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
    E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
    
[...]

PR #605, opened 2017-02-22
Improve performance of history rule

andrew-epstein:
```bash
$ history | wc -l
   26954
$ ll
...
$ fuck
...
DEBUG: Trying rule: history; took: 0:00:02.685889
ls -al [enter/↑/↓/ctrl+c]


$ ll
...
$ fuck
...
DEBUG: Trying rule: history; took: 0:00:00.140865
ls -al [enter/↑/↓/ctrl+c]
```

The issue is the function `get_valid_history_without_current()`; it does a linear scan through the list `executables` once per line of history. Changing `executables` to a set allows for constant-time lookup instead. This should help address #334, #353, and maybe #371.

### DIFF
--- a/thefuck/utils.py
+++ b/thefuck/utils.py
@@ -264,7 +264,7 @@ def get_valid_history_without_current(command):
     from thefuck.shells import shell
     history = shell.get_history()
     tf_alias = get_alias()
-    executables = get_all_executables()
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

PR #606, opened 2017-02-27
Add git_rm_staged rule for removing locally staged changes

josephfrazier:
It would be nice if `thefuck` could help me `git rm` a file I had
already staged. This rule, adapted from `git_rm_local_modifications`,
does that.


nvbn:
Thanks!

PR #607, opened 2017-02-27
Fix aliased `yarn` commands like `yarn ls`

josephfrazier:
[Yarn] has a handful of subcommand [aliases] but does not automatically
[correct] them for the user. This makes it so that `fuck` will do the
trick. For example:

    $ yarn ls
    yarn ls v0.20.3
    error Did you mean `yarn list`?
    info Visit https://yarnpkg.com/en/docs/cli/list for documentation about this command.
    $ fuck
    yarn list [enter/?/?/ctrl+c]

[Yarn]: https://yarnpkg.com/en/
[aliases]: https://github.com/yarnpkg/yarn/blob/0adbc59b18b38b6ac6e4b248e19788ed1d2e80da/src/cli/aliases.js
[correct]: https://github.com/yarnpkg/yarn/pull/1044#issuecomment-253763230

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,7 @@ using the matched rule and runs it. Rules enabled by default are as follows:
 * `vagrant_up` &ndash; starts up the vagrant instance;
 * `whois` &ndash; fixes `whois` command;
 * `workon_doesnt_exists` &ndash; fixes `virtualenvwrapper` env name os suggests to create new.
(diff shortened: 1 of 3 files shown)


nvbn:
Thans!

nvbn:
Maybe we should also handle situations, where `yarn` not suggests us anything, like:
```
✗ yarn whyy
yarn whyy v0.21.3
error Command "whyy" not found.
```

PR #608, opened 2017-03-01
Update PowerShell alias to handle no history

MattKotsenas:
If history is cleared (or the shell is new and there is no history),
invoking thefuck results in an error because the alias attempts to execute
the usage string.

The fix is to check if Get-History returns anything before invoking
thefuck.


nvbn:
Thanks!

PR #609, opened 2017-03-04
Add `yarn_command_not_found` rule

josephfrazier:
This addresses https://github.com/nvbn/thefuck/pull/607#issuecomment-283945505

The code was adapted from the `grunt_task_not_found` rule


nvbn:
Thanks!

PR #612, opened 2017-03-10
Add `yarn_help` rule

josephfrazier:
Yarn likes to keep its documentation online, rather than in `yarn help`
output. For example, `yarn help clean` doesn't tell you anything about
the `clean` subcommand. Instead, it points you towards
https://yarnpkg.com/en/docs/cli/clean

This rule detects when that happens, and suggests opening the URL. <strike>One
caveat is that currently only OSX is supported, as Linux uses `xdg-open`
instead of `open`, and Windows uses `start` (see https://www.dwheeler.com/essays/open-files-urls.html).</strike> EDIT: Added LInux/Windows support


nvbn:
Thanks!

PR #617, opened 2017-03-22
git_stash_pop: Add only updated files

josephfrazier:
This avoids adding untracked files to the repo. See here for a
description of the difference between `git add .` and `git add --update`:

https://stackoverflow.com/questions/572549/difference-between-git-add-a-and-git-add/572660#572660


nvbn:
Thanks!

PR #618, opened 2017-03-22
Created a rule for trying to push a new repository with no commits.

Asday:
I made this because the error message `git` provides is entirely cryptic, and always requires a googling.


nvbn:
Thanks, I've changed the rule a bit.

PR #621, opened 2017-03-26
add rule for: yarn install [pkg]

dfadev:
--- `install` has been replaced with `add` to add new dependencies. Run $0 instead.

https://github.com/yarnpkg/yarn/blob/6e9a9a6596ca8f177f68f6672a1ef4ff16705336/src/reporters/lang/en.js#L18


nvbn:
Thanks!

I've squashed commits and somehow github can't recognize that pull request was merged.

Issue #662, opened 2017-06-27
fuck can't configure itself automatically on bash


Googulator:
Executing "fuck" twice on a system with no alias configured simply displays the configuration message twice, rather than auto-configuring on the second run.

The problem appears to be that _is_second_run() always returns False due to shell.get_history() not getting the complete (recent) history. This is because Bash keeps the history from its current session in memory, and only writes it to .bash_history either when it quits or when "history -a" is called.

The obvious fix would be to call "history -a" in the \_\_init\_\_() of class Bash, but I'm not sure if that's enough.

Googulator:
Filed pull request #663 with a fix. It's not optimal, unfortunately - we would have to call "history -a" in the context of the Bash process that called "fuck", and we can only do so in a subshell, which doesn't have the intended effect.

nvbn:
@Googulator sorry for so huge delay. I've tried a bit simpler approach, now it configures alias if the previous `fuck` was less then 60 seconds ago and from the same shell.

nvbn:
Fixed in 3.20

PR #663, opened 2017-06-27
Make alias autoconfiguration work in Bash

Googulator:
Bash hides its command history from other processes until it exits,
unless the Bash builtin "history -a" is executed. The builtin only
writes the history of the Bash process it has been called in, so there
is no way to programmatically call it from a Python program.

Due to this, the only way to get autoconfiguration to work is to disable
the last history item check for Bash.

### DIFF
--- a/tests/test_not_configured.py
+++ b/tests/test_not_configured.py
@@ -77,7 +77,7 @@ def test_on_first_run(usage_tracker, shell_pid, logs):
 
 def test_on_run_after_other_commands(usage_tracker, shell_pid, shell, logs):
     shell_pid.return_value = 12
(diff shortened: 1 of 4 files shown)


scorphus:
Hello, @Googulator!

First of all, thank you very much for your intent in solving the issue.

As reported in #662, it's been fixed with a different approach, which you can check [here](6a096155dcbb4b252cd486f32a7ac52a0aff2850). So, I believe we can move on and close this pull request. What are your thoughts?

Thanks.

scorphus:
@Googulator, I'll move on and close this pull request. Please feel free to reopen it – or submit a new one – if you see fit.

PR #669, opened 2017-07-18
#630 Catching the escaped space in filenames

tomoshi0809:
This addresses  #630 

The escaped space ("\ ") in the filename which a command references is interpreted a normal space (" ").
I suggest thefuck should catch the escaped space ("\ ").
(I'm afraid this suggestion conflicts with the current correcting policy of "tar" and "zip")


nvbn:
Thanks!

PR #683, opened 2017-08-25
simplify fuck-it alias

rdpate:
added benefit: only affects single command rather than modifying current environment


nvbn:
It was a workaround for older version of thefuck, now we have `-y` option that I forgot to mention in the readme. So now there's no need for `fuck-it` or similar alias.

PR #688, opened 2017-08-31
Update README.md

Nfinished:
😉


josephfrazier:
Sorry, but I'm not a fan of this one, personally. There are loads of places we could write "fuck" (or variations thereof, it's a versatile word!), and I don't really want to start adding it unnecessarily.

PR #696, opened 2017-09-26
Support parameters in PowerShell

MattKotsenas:
Update the PowerShell alias so it passes thefuck parameters (e.g. `-y` or `-r`).


nvbn:
Thanks!

PR #698, opened 2017-10-05
Basic DNF support

Epse:
This PR contains basic support for `DNF` the Fedora package manager.

It contains a new specific, which only contains a variable `dnf_available`, very similar to `thefuck.specific.apt.apt_available`.

Also there is a `dnf_no_such_command` rule that parses `DNF`'s help output for commands and autofixes your input. It is enabled based on the `dnf_available` value.

`flake8` had no complaints about my code, but if you do, I'd welcome the advice!

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,7 @@ Enabled by default only on specific platforms:
 * `brew_uninstall` &ndash; adds `--force` to `brew uninstall` if multiple versions were installed;
 * `brew_unknown_command` &ndash; fixes wrong brew commands, for example `brew docto/brew doctor`;
 * `brew_update_formula` &ndash; turns `brew update <formula>` into `brew upgrade <formula>`;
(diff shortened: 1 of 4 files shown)


Epse:
It apears AppVeyor is broken...

scorphus:
Nice! This one also lacks a proper test and a line in README.md explaining the new rule. Can you please add both? Thanks!

Epse:
I assumed I shouldn't touch the README. Oh well, I'll fix it

scorphus:
Looking good.

How about tests?

Epse:
On their way. Running into a couple of issues where tests are failing but the same deal actually typed and fucked in a terminal works... Seems to be due to byte strings and regular strings

Epse:
There are now tests!

nvbn:
Thanks!

Issue #655, opened 2017-06-05
php, fuck !


Nek-:
Hey

I tried "fuck" after `php -s localhost:8000`, it should give me `php -S localhost:8000` but I got `php -a localhost:8000` (which is executing code instead of running server)

scorphus:
There's no such rule, @Nek-, unfortunately. Are you willing to code it?

Epse:
made a quick PR

PR #702, opened 2017-10-05
quick fix for #655

Epse:
Simply autocorrects `php -s` to `php -S` as per #655

### DIFF
--- a/thefuck/rules/php_s.py
+++ b/thefuck/rules/php_s.py
@@ -1,9 +1,14 @@
-from thefuck.utils import replace_argument
+from thefuck.utils import replace_argument, for_app
 
 
+@for_app('php')
 def match(command):
     return "php -s" in command.script
 
 
 def get_new_command(command):
     return replace_argument(command.script, "-s", "-S")
+
+
+enabled_by_default = True
+requires_output = False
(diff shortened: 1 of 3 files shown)


thefuck/rules/php_s.py
```diff
@@ -0,0 +1,9 @@
+from thefuck.utils import replace_argument
+
+
+def match(command):
+    return "php -s" in command.script
+
+
+def get_new_command(command):
+    return replace_argument(command.script, "-s", "-S")
```
scorphus: `-s` (lowercase) is a valid command and will timeout when ran by TheFuck. Meaning output will be empty. The rule must tell it doesn't require output. So, add this to the end of the rule:

```python
requires_output = False
```

Epse:
is this one okay now?

thefuck/rules/php_s.py
```diff
@@ -0,0 +1,14 @@
...
+@for_app('php')
+def match(command):
+    return "php -s" in command.script
+
+
+def get_new_command(command):
+    return replace_argument(command.script, "-s", "-S")
+
+
+enabled_by_default = True
```
scorphus: This line can be omitted.

Epse:
This should be okay now then.

nvbn:
Thanks!

PR #705, opened 2017-10-06
#N/A: Use curl to download get-pip.py on AppVeyor

scorphus:
`curl` seems to be smarter than `net.webclient` when dealing with SSL.


nvbn:
Thanks!

PR #706, opened 2017-10-08
Fix minor issues with git_remote_delete

scorphus:
* Do not require `git` in the script
* Replace the first single occurrence of `delete`


nvbn:
Thanks!

PR #709, opened 2017-10-10
#702: fix minor issues with `php_s` rule

scorphus:
Unfortunately, I didn't catch these issues while reviewing #702.

After looking more closely at `php` options, `-S` requires additional arguments (&lt;address&gt;:&lt;port&gt;) and `-s` may produce output if used that way. So, matching ` -s ` seems to be better.

Also, `@for_app('php')` already asserts the presence of `php ` in the command script. Matching `php -s` prevents the rule from fixing commands like `php -t public -s 0.0.0.0:8080`.

Reviews are more than welcome 😃

### DIFF
--- a/tests/rules/test_php_s.py
+++ b/tests/rules/test_php_s.py
@@ -3,8 +3,12 @@ from thefuck.rules.php_s import get_new_command, match
 from thefuck.types import Command
 
 
-def test_match():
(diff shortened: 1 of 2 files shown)


nvbn:
Thanks!
I've made `match` a bit more strict.

scorphus:
> I've made `match` a bit more strict

Great, I like it! Thanks!

PR #710, opened 2017-10-12
Bugfix: brew_update_formula.py

matthewfallshaw:
Sample command output is:

    Error: This command updates brew itself, and does not take formula names.
    Use 'brew upgrade thefuck' instead.

This will never match the previous `"Use 'brew upgrade <formula>'" in command.output` test.


nvbn:
Thanks!

PR #722, opened 2017-10-31
git_flag_after_filename: Handle new error message

josephfrazier:
git version 2.15.0 has updated wording that we need to account for.

See https://github.com/git/git/commit/2a5aa826eec2eec64b4065f599ee7fb613a871ba


scorphus:
Thank you, @josephfrazier

Issue #728, opened 2017-11-06
`fuck` should suggest specific heroku apps when there are multiple


josephfrazier:
Here's an example of current behavior, as of https://github.com/nvbn/thefuck/commit/8fb5ddefb60c59f8bd90a72698c1a384f1433c5e:

```
$ heroku pg
 ▸    Multiple apps in git remotes
 ▸    Usage: --remote heroku-dev
 ▸    or: --app myapp-dev
 ▸    Your local git repository has more than 1 app referenced in git remotes.
 ▸    Because of this, we can't determine which app you want to run this command against.
 ▸    Specify the app you want with --app or --remote.
 ▸    Heroku remotes in repo:
 ▸    myapp (heroku)
 ▸    myapp-dev (heroku-dev)
 ▸    
 ▸    https://devcenter.heroku.com/articles/multiple-environments
$ fuck
heroku open [enter/↑/↓/ctrl+c]
```

Here's an example of behavior that would be more useful:

```
$ heroku pg
 ▸    Multiple apps in git remotes
 ▸    Usage: --remote heroku-dev
 ▸    or: --app myapp-dev
 ▸    Your local git repository has more than 1 app referenced in git remotes.
 ▸    Because of this, we can't determine which app you want to run this command against.
 ▸    Specify the app you want with --app or --remote.
 ▸    Heroku remotes in repo:
 ▸    myapp (heroku)
 ▸    myapp-dev (heroku-dev)
 ▸    
 ▸    https://devcenter.heroku.com/articles/multiple-environments
$ fuck
heroku pg --app myapp-dev [enter/↑/↓/ctrl+c]
```

It would be even better if all of the apps were suggested separately:

```
heroku pg --app myapp-dev
heroku pg --app myapp
```

These can be derived from the initial error message, in the lines following ` ▸    Heroku remotes in repo:`

---

I'll try to make time to submit a patch, but I wanted to go ahead and create an issue for it.

PR #729, opened 2017-11-07
Add heroku_multiple_apps rule

josephfrazier:
Closes https://github.com/nvbn/thefuck/issues/728


josephfrazier:
Thanks for the review @scorphus! I was also thinking they should be squashed together, so I went ahead and did it just now :)

Issue #725, opened 2017-11-04
Execting fuck prints thefuck usage 


agross:
```sh
$ zsh --version
zsh 5.4.2 (x86_64-apple-darwin16.7.0)

agross@firiel ~
$ thefuck --version
The Fuck 3.24 using Python 3.6.3

agross@firiel ~
$ git stat
git: 'stat' is not a git command. See 'git --help'.

The most similar commands are
	status
	stage
	stash

agross@firiel ~
$ fuck
usage: thefuck [-v] [-a [ALIAS]] [-l SHELL_LOGGER]
               [--enable-experimental-instant-mode] [-h] [-y | -r] [-d]
               [command [command ...]]

agross@firiel ~
$ whence -f fuck
fuck () {
	TF_HISTORY=$(fc -ln -10)
	TF_CMD=$(
                    TF_ALIAS=fuck
                    TF_SHELL_ALIASES=$(alias)
                    TF_HISTORY=$TF_HISTORY
                    PYTHONIOENCODING=utf-8
                    thefuck THEFUCK_ARGUMENT_PLACEHOLDER $*
                )  && eval $TF_CMD
	test -n "$TF_CMD" && print -s $TF_CMD
}

agross@firiel ~
$ setopt
alwaystoend
autocd
noautoparamslash
autopushd
nobgnice
completeinword
correct
correctall
extendedhistory
globdots
histexpiredupsfirst
histignoredups
histreduceblanks
histverify
incappendhistory
interactive
interactivecomments
login
longlistjobs
menucomplete
monitor
nonotify
promptsubst
pushdignoredups
recexact
rmstarwait
sharehistory
shinstdin
zle
```

cojack:
```
╭─cojack@cojack-laptop ~ ‹v8.5.0› 
╰─$ git bara                                                                                                                                                                                                                                                         20:45:22
git: 'bara' is not a git command. See 'git --help'.

The most similar command is
        branch
╭─cojack@cojack-laptop ~ ‹v8.5.0› 
╰─[1 ↵]─$ fuck                                                                                                                                                                                                                                                       20:45:25
[WARN] Script not found in output log
No fucks given
```
still doesn't work

agross:
@cojack What do you mean by it "still" doesn't work?

akuma:
+1

ianmurrays:
Same behavior here, it happened to me after upgrading thefuck to the latest version. Unfortunately I'm not sure what my previous version was.

[...]

PR #733, opened 2017-11-16
Fix zsh alias

MasterNayru:
It would appear that the alias printed out in v3.24 has a variable scoping issue with the TF_HISTORY variable. The bash alias exports TF_HISTORY and does an unset on it afterwards, which seems to make it work fine. With the zsh alias in its current form, os.environ never sees the TF_HISTORY variable, and since $TF_PREVIOUS is no longer being fed into thefuck it just prints out the usage message. I have tested this change locally and it seems to make it work again.

This should resolve issue #725


nvbn:
Thanks!

Issue #738, opened 2017-11-24
Don't rely on $SHELL for detecting shell


nvbn:
When, for example, bash run on top of  zsh, `$SHELL` will be zsh and thefuck will generate wrong alias.

We use `$SHELL` for performance reason, so for fixing this we need to:
* get shell from parent pids in `thefuck --alias`;
* store it in `$TF_SHELL`;
* use `$TF_SHELL` in `fuck`.

Related issues: #717 #731 #735

EmberQuill:
Was going to report that I'm getting an Unexpected End of File when starting bash from zsh but this ended up being the cause of the issue. `eval $(SHELL=bash thefuck--alias)` worked.

PR #744, opened 2017-12-17
Continue work started in #738

scorphus:
Tests are failing because, [since 3.3.0](https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23), pytest no longer supports Python 3.3. Please, stand by for another pull request that discusses ways to solve this.

### DIFF
--- a/tests/shells/test_bash.py
+++ b/tests/shells/test_bash.py
@@ -51,6 +51,7 @@ class TestBash(object):
     def test_app_alias_variables_correctly_set(self, shell):
         alias = shell.app_alias('fuck')
         assert "fuck () {" in alias
+        assert 'TF_SHELL=bash' in alias
         assert "TF_ALIAS=fuck" in alias
         assert 'PYTHONIOENCODING=utf-8' in alias
         assert 'TF_SHELL_ALIASES=$(alias)' in alias
--- a/tests/shells/test_fish.py
+++ b/tests/shells/test_fish.py
@@ -71,6 +71,7 @@ class TestFish(object):
         assert 'function fuck' in shell.app_alias('fuck')
         assert 'function FUCK' in shell.app_alias('FUCK')
         assert 'thefuck' in shell.app_alias('fuck')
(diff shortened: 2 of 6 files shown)


nvbn:
Do we need to support Python 3.3?

josephfrazier:
I'm personally okay with dropping Python 3.3 support, but I pushed https://github.com/nvbn/thefuck/commit/f966ecd4f5b8221ee15e843f5ec287e1f7cca940 to get the tests working while we decide on that.

scorphus:
Well, I guess we should move forward. If pytest made the move first, they did so for a reason. And even it being a test library, I think we're better dropping Python 3.3 support instead of freezing `pytest<3.3`. But that's MHO.

scorphus:
> I pushed f966ecd to get the tests working

Oh, I guess you accidentally pushed to master, @josephfrazier.

scorphus:
Also, according to [Python 3.3.7 release notes](https://www.python.org/downloads/release/python-337/), _"Python 3.3.x has reached end-of-life"_ and _"all support for the 3.3 series of releases ended on 2017-09-29, five years after the initial release. This release, Python 3.3.7, was the final release for the 3.3 series."_

josephfrazier:
Yeah, I agree we should drop Python 3.3 and unfreeze `pytest`. I just pushed https://github.com/nvbn/thefuck/commit/f966ecd4f5b8221ee15e843f5ec287e1f7cca940 to master so that I could unblock myself on https://github.com/nvbn/thefuck/pull/745, and to ensure that any new PRs wouldn't fail for no fault of their own (like this one did), just in case there was any debate over dropping Python 3.3.

It sounds like the three of us are in agreement though, so I'll open a new PR reverting f966ecd4f5b8221ee15e843f5ec287e1f7cca940 and dropping Python 3.3 from the tests.

scorphus:
Thanks @josephfrazier for dropping Py33 support. This one looks all good now, I'll go ahead and merge it.

PR #751, opened 2017-12-31
Don't add duplicate remote/branch name

davidhart82:
@josephfrazier this should cover all of the corner cases for #745


thefuck/rules/git_push.py
```diff
@@ -32,10 +32,12 @@ def get_new_command(command):
         # In case of `git push -u` we don't have next argument:
         if len(command_parts) > upstream_option_index:
             command_parts.pop(upstream_option_index)
-    # If the only argument is the remote/branch, remove it.
-    # git's suggestion includes it, so it won't be lost, and we would duplicate it otherwise.
-    elif len(command_parts) is 3 and command_parts[2][0] != '-':
-        command_parts.pop(2)
+    elif len(command_parts) > 2:
```
davidhart82: Not sure if this is valuable; or whether it should be something like:
`elif len(command_parts) > command_parts.index('push')`

davidhart82: Actually, probably more like:

`elif len(command_parts) > command_parts.index('push') + 1:`

josephfrazier:
Thanks!

PR #753, opened 2018-01-01
Fix fish shell aliasing

### LINKED ISSUE #727: Parsing Fish shell aliases seems to be broken
In `~/.config/fish/conf.d/aliases.fish` I have `alias g 'git'` (a common alias, I think, for `g -> git`). This alias works, and even Fish auto-completion recognizes it and works.

I'm running `fish, version 2.6.0` with `The Fuck 3.24 using Python 3.6.3`, and just updated `~/.config/fish/functions/fuck.fish` to be:

```
function fuck -d "Correct your previous console command"
  set -l fucked_up_command $history[1]
  env TF_ALIAS=fuck PYTHONIOENCODING=utf-8 thefuck $fucked_up_command | read -l unfucked_command
  if [ "$unfucked_command" != "" ]
    eval $unfucked_command
    builtin history delete --exact --case-sensitive -- $fucked_up_command
    builtin history merge ^ /dev/null
  end
end
```

The `fuck` alias works for me, except, it's not resolving aliases!

I added a new rule to `~/.config/thefuck/rules/git_worktree_checkout.py`:

```
import re
from thefuck.specific.git import git_support

@git_support
def match(command):
    return ('checkout' in command.script
            and 'fatal:' in command.output
            and 'is already checked out at' in command.output)

@git_support
def get_new_command(command):
    directory_name = re.findall(
        r"fatal: '[^']+' is already checked out at '([^']+)'" , command.output)[0]
    return 'cd {}'.format(directory_name)
```

To fix the scenario of checking out a branch that's already checked out in another Git worktree. Using the `@git_support`, my _Git aliases_ (i.e. `co -> checkout`) are expanded properly, but Fish gives me `No fucks given` if I try to use `g` instead of `git`.

```
$ pwd
/home/andschwa/src/mesos
$ git checkout master
fatal: 'master' is already checked out at '/home/andschwa/src/mesos-master'
$ fuck
cd /home/andschwa/src/mesos-master [enter/↑/↓/ctrl+c]
$ cd ~/src/mesos
$ git co master
fatal: 'master' is already checked out at '/home/andschwa/src/mesos-master'
$ fuck
cd /home/andschwa/src/mesos-master [enter/↑/↓/ctrl+c]
$ cd ~/src/mesos
$ g checkout master
fatal: 'master' is already checked out at '/home/andschwa/src/mesos-master'
$ fuck
No fucks given
```

It _looks_ to me that this scenario is [supposed to be supported](https://github.com/nvbn/thefuck/blob/8fb5ddefb60c59f8bd90a72698c1a384f1433c5e/thefuck/shells/fish.py#L14), but I cannot get it to work. The `g` alias is in the output of `fish -ic functions`:

```
., N_, abbr, alias, cd, contains_seq, delete-or-exit, dirh, dirs, down-or-search, e, ec, edit_command_buffer, eval, export, extract, fish_clipboard_copy,
fish_clipboard_paste, fish_config, fish_default_key_bindings, fish_default_mode_prompt, fish_fallback_prompt, fish_greeting, fish_hybrid_key_bindings,
fish_indent, fish_key_reader, fish_md5, fish_mode_prompt, fish_prompt, fish_sigtrap_handler, fish_title, fish_update_completions, fish_vi_cursor,
fish_vi_key_bindings, fish_vi_mode, fuck, funced, funcsave, g, grep, help, history, hostname, isatty, la, ll, ls, man, math, nextd, nextd-or-forward-word,
open, p, package, popd, prevd, prevd-or-backward-word, prompt_hostname, prompt_pwd, psub, pushd, rbenv, rbt_post, rbt_update, realpath, rg, seq, setenv,
string, suspend, trap, type, umask, up-or-search, vared,
```

I have tried:

* updating thefuck
* switching to the Python 3.6 version of thefuck
* mucking around with the `THEFUCK_OVERRIDDEN_ALIASES` but learned from another issue that is for the scenario of `git` being aliased to something else (e.g. `hub`)
* restarting my shells etc. entirely
* ensuring `fish` is my login shell with `chsh`

Issue #479 seems to be the same problem, but it went away with a newer version of Python(?) I guess.

davidhart82:
This should fix #727.

### DIFF
--- a/tests/shells/test_fish.py
+++ b/tests/shells/test_fish.py
@@ -43,7 +43,8 @@ class TestFish(object):
         ('open', 'open'),
         ('vim', 'vim'),
         ('ll', 'fish -ic "ll"'),
-        ('ls', 'ls')])  # Fish has no aliases but functions
+        ('ls', 'ls'),
+        ('g', 'git')])
     def test_from_shell(self, before, after, shell):
         assert shell.from_shell(before) == after
 


andschwa:
Thanks @davidhart82! I applied this to master and tested, it fixes my scenario!

josephfrazier:
Thanks for the fix @davidhart82 (and for testing it, @andschwa)! This looks good to me, but I'll give a little more time for the other maintainers to have a look, if they wish. I don't use fish myself, so I'm not as familiar with this area of the code.

davidhart82:
Yes, I'm wondering about a scenario where a user aliases a fish function, I should probably check that the aliased command is also not in the dict of aliases.

scorphus:
Thanks @davidhart82 for the fix, @andschwa for testing it and @josephfrazier for allowing another maintainer (and fish user) to have a look :+1: I'll do so

scorphus:
Very good job, @davidhart82! Thank you!

> Yes, I'm wondering about a scenario where a user aliases a fish function, I should probably check that the aliased command is also not in the dict of aliases.

In this scenario, the `overridden` set [should already contain](https://github.com/nvbn/thefuck/wiki/Shell-aliases#fish) the _aliased_ function. :smiley:

PR #754, opened 2018-01-01
Suggest git checkout -b

### LINKED ISSUE #632: git checkout new branch suggestion improvement
Currently if i do the following:

    > git checkout my-new-branch
     error: pathspec 'my-new-branch' did not match any file(s) known to git.
    > fuck
I get the following suggestion:

    > git branch my-new-branch && git checkout my-new-branch

This works fine, but what i more likely meant (and this does the same thing) is:

    > git checkout -b my-new-branch

This is much more succinct and should probably be the suggestion in my opinion.

davidhart82:
This should fix #632.

While playing with this fix, it wasn't obvious to me why git_checkout.py is suggesting changes for `git commit`; the suggestions don't seem sensible.

### DIFF
--- a/tests/rules/test_git_checkout.py
+++ b/tests/rules/test_git_checkout.py
@@ -52,7 +52,7 @@ def test_get_branches(branches, branch_list, git_branch):
 @pytest.mark.parametrize('branches, command, new_command', [
     (b'',
      Command('git checkout unknown', did_not_match('unknown')),
-     'git branch unknown && git checkout unknown'),
+     'git checkout -b unknown'),
     (b'',
      Command('git commit unknown', did_not_match('unknown')),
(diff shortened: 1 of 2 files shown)


josephfrazier:
Thanks @davidhart82! I just made a slight tweak (and probably should do the same for other rules).

davidhart82:
Hmm...yeah, that whole thing doesn't seem very robust. What if I wanted `git - -git_dir=xxx checkout branch`. I think a lot of the suggestions don't cater for this when using positional references in script_parts.

josephfrazier:
Yeah, I guess it would be better to filter out the `script_parts` that start with `-` before checking the one at a particular position. I'm not sure if this has been an issue in practice, but if you'd like to implement it, feel free to submit a PR.

PR #760, opened 2018-01-02
Escape single quote in branch names

### LINKED ISSUE #600: Error when correcting `git push` to branch with quotes in name from fish
When called from fish, using thefuck to correct a missing upstream branch fails if the branch name contains quotes or special characters.

Example below,
```
~/my-app> git push
fatal: The current branch feat/let's-do-this has no upstream branch.

    git push --set-upstream origin feat/let's-do-this

~/my-app> fuck
git push --set-upstream origin feat/let's-do-this [enter/↑/↓/ctrl+c]
Unexpected end of string, quotes are not balanced
- (line 1): begin; git push --set-upstream origin feat/let's-do-this
                                                          ^
from sourcing file -
	called on line 60 of file /usr/local/Cellar/fish/2.4.0/share/fish/functions/eval.fish

in function 'eval'
	called on line 1 of file -

in function 'fuck'
	called on standard input

source: Error while reading file '-'
```

davidhart82:
This should fix #600.

### DIFF
--- a/tests/rules/test_git_push.py
+++ b/tests/rules/test_git_push.py
@@ -3,33 +3,33 @@ from thefuck.rules.git_push import match, get_new_command
 from thefuck.types import Command
 
 
-@pytest.fixture(scope='function')
-def output(request):
-    if not request.param:
+@pytest.fixture
+def output(branch_name):
+    if not branch_name:
         return ''
     return '''fatal: The current branch {} has no upstream branch.
 To push the current branch and set the remote as upstream, use
 
     git push --set-upstream origin {}
 
-'''.format(request.param, request.param)
+'''.format(branch_name, branch_name)
 
 
-@pytest.mark.parametrize('script, output', [
+@pytest.mark.parametrize('script, branch_name', [
     ('git push', 'master'),
-    ('git push origin', 'master')], indirect=['output'])
-def test_match(script, output):
+    ('git push origin', 'master')])
+def test_match(output, script, branch_name):
     assert match(Command(script, output))
 
 
-@pytest.mark.parametrize('script, output', [
+@pytest.mark.parametrize('script, branch_name', [
     ('git push master', None),
-    ('ls', 'master')], indirect=['output'])
-def test_not_match(script, output):
+    ('ls', 'master')])
+def test_not_match(output, script, branch_name):
     assert not match(Command(script, output))
 
 
-@pytest.mark.parametrize('script, output, new_command', [
+@pytest.mark.parametrize('script, branch_name, new_command', [
     ('git push master', 'master',
      'git push --set-upstream origin master'),
     ('git push master', 'master',
@@ -51,6 +51,6 @@ def test_not_match(script, output):
     ('git -c test=test push --quiet origin', 'master',
      'git -c test=test push --set-upstream origin master --quiet'),
     ('git push', "test's",
-     "git push --set-upstream origin test\\'s")], indirect=['output'])
-def test_get_new_command(script, output, new_command):
+     "git push --set-upstream origin test\\'s")])
+def test_get_new_command(output, script, branch_name, new_command):
     assert get_new_command(Command(script, output)) == new_command
--- a/tests/rules/test_git_push.py
+++ b/tests/rules/test_git_push.py
@@ -30,9 +30,7 @@ def test_not_match(output, script, branch_name):
 
 
 @pytest.mark.parametrize('script, branch_name, new_command', [
-    ('git push master', 'master',
-     'git push --set-upstream origin master'),
-    ('git push master', 'master',
+    ('git push', 'master',
      'git push --set-upstream origin master'),
     ('git push master', 'master',
      'git push --set-upstream origin master'),
--- a/tests/rules/test_git_push.py
+++ b/tests/rules/test_git_push.py
@@ -4,38 +4,51 @@ from thefuck.types import Command
 
 
 @pytest.fixture
-def output():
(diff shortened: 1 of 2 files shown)


thefuck/rules/git_push.py
```diff
@@ -39,6 +39,6 @@ def get_new_command(command):
         while len(command_parts) > push_idx and command_parts[len(command_parts) - 1][0] != '-':
             command_parts.pop(len(command_parts) - 1)
 
-    arguments = re.findall(r'git push (.*)', command.output)[0].strip()
+    arguments = re.findall(r'git push (.*)', command.output)[0].replace("'", r"\'").strip()
```
josephfrazier: Instead of replacing just single quotes, I think we should use [`shell.quote`](https://github.com/nvbn/thefuck/blob/83e1710712a9b48fee1822109a10a8324fe64930/thefuck/shells/generic.py#L101-L109) here.

tests/rules/test_git_push.py
```diff
@@ -4,38 +4,51 @@
...
-    assert get_new_command(Command('git -c test=test push --quiet origin', output))\
-        == "git -c test=test push --set-upstream origin master --quiet"
+    git push --set-upstream origin {}
+
+'''.format(branch, branch)
+
+
+@pytest.mark.parametrize('command', [
+    Command('git push', output('master')),
+    Command('git push master', output('master'))])
```
scorphus: This way `output` is not being used as a fixture. I'll comment on that ASAP.

thefuck/rules/git_push.py
davidhart82: shell.quote uses shlex.quote which escapes the whole string; so you end up with something like:
`'--set-upstream origin test'"'"'s'`

tests/rules/test_git_push.py
davidhart82: It's identical in use to `test_git_checkout.py`.

scorphus: Please consider the following:

```python
@pytest.fixture
def branch():
    return 'master'


# output fixture definition stays the same


@pytest.mark.parametrize('script', [
    'git push',
    'git push master'])
def test_match(script, output):
    assert match(Command(script, output))


@pytest.mark.parametrize('script, branch', [
    ('git push', 'it\'s-just-a-branch'),
    ('git push let\'s-push-it', 'let\'s-push-it')])
def test_match_branch(script, branch, output):
    assert match(Command(script, output))


# the rest must be adapted
```

What do you guys think of this change?

thefuck/rules/git_push.py
josephfrazier: Oh, I had overlooked the fact that the command contains other arguments as well. Thanks for pointing that out. This seems fine to me now.

davidhart82: Yes, I had hoped there was something better, and I had a quick search, but nothing jumped out to me.  I do agree though, that this is kind of a sticking-plaster.  I do find it weird that Git allows you to use quotes in branch names...

tests/rules/test_git_push.py
davidhart82: Sure, I can give that a go.

davidhart82: It's not obvious to me how a) your branch test fixture is used and b) what you expect the branch in test_match_branch to be used for?

davidhart82: No worries, I think I've managed to sort it.  What do you think?

tests/rules/test_git_push.py
```diff
@@ -3,39 +3,55 @@
...
-        == "git push --set-upstream origin master"
-    assert get_new_command(Command('git push --quiet', output))\
-        == "git push --set-upstream origin master --quiet"
-    assert get_new_command(Command('git push --quiet origin', output))\
-        == "git push --set-upstream origin master --quiet"
-    assert get_new_command(Command('git -c test=test push --quiet origin', output))\
-        == "git -c test=test push --set-upstream origin master --quiet"
+@pytest.fixture(scope='function')
+def output(request):
+    if not request.param:
```
scorphus: You just read my mind! I was just writing a comment suggesting indirect parameters. Top notch! :clap:

tests/rules/test_git_push.py
```diff
@@ -3,39 +3,55 @@
...
+        return ''
+    else:
```
scorphus: This else can be removed :wink:

scorphus: This else can be removed :wink:

scorphus:
Yeah... there's `'master'` all around but I think that's okay as the tests are more readable now.

scorphus:
Your on :fire: @davidhart82! Let us know when you're done :slightly_smiling_face:

davidhart82:
If you're worried about `master` being everywhere, could almost certainly set `branch` to have a default argument in `output`.

davidhart82:
@scorphus all done ;)

tests/rules/test_git_push.py
```diff
@@ -4,38 +4,53 @@
...
+    assert not match(Command(script, output))
+
+
+@pytest.mark.parametrize('script, branch_name, new_command', [
+    ('git push master', 'master',
+     'git push --set-upstream origin master'),
+    ('git push master', 'master',
+     'git push --set-upstream origin master'),
+    ('git push master', 'master',
+     'git push --set-upstream origin master'),
```
josephfrazier: This test and the one above are duplicates of the previous one. Other than that, this looks good to me!

PR #761, opened 2018-01-03
Add apt_upgrade rule

josephfrazier:
This suggests `apt upgrade` after `apt list --upgradable` if there are
packages to upgrade. It pairs well with the `apt_list_upgradable` rule,
which suggests `apt list --upgradable` after `apt update` if there are
packages to upgrade.


nvbn:
Thanks!

PR #765, opened 2018-01-03
Add rule for ADB unknown commands

davidhart82:
New feature; fix misspelled Android ADB commands.

### DIFF
--- a/thefuck/rules/adb_unknown_command.py
+++ b/thefuck/rules/adb_unknown_command.py
@@ -2,16 +2,29 @@ from thefuck.utils import is_app, get_closest, replace_argument
 
 
 # the most common ADB commands
-_ADB_COMMANDS = [
+_ADB_COMMANDS = (
+    'backup',
+    'bugreport',
+    'connect',
     'devices',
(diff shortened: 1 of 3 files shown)


thefuck/rules/adb_unknown_command.py
```diff
@@ -0,0 +1,28 @@
+from thefuck.utils import is_app, get_closest, replace_argument
+
+
+# the most common ADB commands
+_ADB_COMMANDS = [
```
davidhart82: There are more, do we want them all?

josephfrazier: Yeah, I think we may as well go ahead and list them all.

thefuck/rules/adb_unknown_command.py
```diff
@@ -0,0 +1,41 @@
+from thefuck.utils import is_app, get_closest, replace_argument
+
+
+# the most common ADB commands
```
scorphus: They are now *ALL* AMD commands :tada:

PR #764, opened 2018-01-03
Offer git commit --amend after previous git commit

davidhart82:
Added proposal for feature request #664 (I use this a lot at work!).

PR #766, opened 2018-01-05
Test getting status code

davidhart82:
This, or something very similar, was what I hoped existed when I was implementing the `git commit - - amend` rule.

Do we think this, or a variant of it, might be useful? Before I do any more work on it.


josephfrazier:
I was thinking the same thing when I saw https://github.com/nvbn/thefuck/pull/764#discussion_r159547506. I think it'd be useful, and would definitely welcome further work! I just merged #764, so this PR can build off of the new master and use `git_commit_amend` as the first rule to be tested.

It would also be nice to have functional tests that ensure the status code is actually being set properly, since the rule tests won't cover that.

scorphus:
Great! I also thought about this some while ago, I just can't remember what rule made me think of it, but there are some other rules that might benefit from this as well.

nvbn:
Sorry for slow response, it's great idea.

davidhart82:
I'll revive this now I'm stuck at home; will look at these functional tests and try to decipher how they work.

Issue #469, opened 2016-02-29
Change way of interaction with shells


nvbn:
I guess it would be nice to:
- put shell aliases in environment variable (`TF_SHELL_ALIASES`) in alias and get it from there, instead of calling shell from app (done for bash and zsh);
- move all writes to history to alias (done for bash and zsh).

It would speedup thefuck, and probably will fix #428, #422 and #402.

bhavybarca:
@nvbn i would like to help.
P.s. i am a newbie

PR #767, opened 2018-01-06
Make fish like bash/zsh

davidhart82:
This should implement the request made in #469 (which has already been completed for bash/zsh) for fish.

### DIFF
--- a/tests/shells/test_fish.py
+++ b/tests/shells/test_fish.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import os
 import pytest
 from thefuck.shells import Fish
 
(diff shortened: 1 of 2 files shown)


scorphus:
:thinking: I have a branch in my local repo (and maybe remote too) with attempts to cover #469. Ranging from this you're proposing to a very different and more complex solution. All to my dislike mainly because they made `fuck` a lot slower. Specially on my macbook with so many plugins.

davidhart82:
Interesting.  My timings suggest it's quicker, but then I don't have many plugins installed.  My setup at work is a touch more mature than at home, but even then, I don't really have an exceptionally complex setup.

scorphus:
I'll work on this asap.

PR #770, opened 2018-01-08
Add installation instructions on FreeBSD

yzgyyang:
misc/thefuck has recently been committed to the FreeBSD ports tree (https://svnweb.freebsd.org/ports?view=revision&revision=458123).


nvbn:
Thanks!

Issue #784, opened 2018-02-16
Absolute paths not recognized (?) by `cd_correction`


alexbarcelo:
thefuck version:

    The Fuck 3.25 using Python 3.6.3

Using `fish` in Ubuntu 17.04

To reproduce the bug:

```bash
cd /var/log/apr
fuck  # doesn't recommend /var/lib/apt as the correct path

cd /var/lib
cd apr
fuck  # recommends /var/lib/apt as the correct path
```

Both commands with DEBUG activated yields a lot of lines finishing with:

`DEBUG: Trying rule: cd_correction; took: 0:00:00.000028`

So it seems that `cd_correction` is behaving differently when the path is absolute.

PR #785, opened 2018-02-16
Absolute path feature for cd_correction

alexbarcelo:
This is motivated by #784 

According to my understandment of the code, this should fix the scenario of "`dest` starts with `/`". However, I had some problems with local tests and still I am not sure if I am overlooking some corner case.

### DIFF
--- a/tests/functional/test_cd_correction.py
+++ b/tests/functional/test_cd_correction.py
@@ -1,0 +1,40 @@
+import pytest
+
+dockerfile = u'''
+FROM python:3
+RUN adduser --disabled-password --gecos '' test
+WORKDIR /src
+USER test
+RUN echo 'eval $(thefuck --alias)' > /home/test/.bashrc
+RUN echo > /home/test/.bash_history
+RUN mkdir -p /home/test/some/random/folder
+RUN mkdir -p /home/test/another/random/folder
+USER root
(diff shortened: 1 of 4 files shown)


scorphus:
How willing are you to write tests for this rule?

alexbarcelo:
Can do.

This WE I suppose I will complete the PR then.

alexbarcelo:
There are no unittest for the correction, but that needs filesystem so... not very unittest-able I'm afraid.

@scorphus, are the provided basic tests ok? Do you feel that the new functional is enough for the cd_correction? `coveralls` doesn't detect the executed code in functional tests as covered (AFAIK), but the functional should indeed be testing that.

alexbarcelo:
There are no unittest for the correction, but that needs filesystem so... not very unittest-able I'm afraid.

@scorphus, are the provided basic tests ok? Do you feel that the new functional is enough for the cd_correction? `coveralls` doesn't detect the executed code in functional tests as covered (AFAIK), but the functional should indeed be testing that.

nvbn:
Thanks!

We don't actually use functional tests for rules, because it's too slow, so I removed it.

nvbn:
I squashed commits manually, it's not just closed, but merged =)

Issue #772, opened 2018-01-11
Colors messed up after any 'fuck' command


hugo-vrijswijk:
The output of `thefuck --version`:

    The Fuck 3.25 using Python 3.6.4

Your shell and its version:

    PowerShell v5.1.16299.98

Your system:

    Windows 10 (1709)

How to reproduce the bug:

Using any fuck command in PowerShell messes up the colors. See image. Calling thefuck is fine
![image](https://user-images.githubusercontent.com/10114577/34844415-af1b9e90-f711-11e7-9ea5-76943b2f46e3.png)

These are my debug settings: 

```
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': False,
 'no_colors': False,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': WindowsPath('C:/Users/hugod/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
```

And I have set `$env:PYTHONIOENCODING = "utf-8"` in my `$PROFILE`

Happens in both ConEmu, Windows console host and in Administrator sessions of both. Hope this is enough information for a start

MattKotsenas:
I can repro this as well. If thefuck changes the console color via `no fucks given`, or `ctrl + c`, it doesn't change back.

I'd be interested in taking a look at this, since I originally added PowerShell support (and created this mess ;)), but I'm not sure where the tests for this should go. Can someone suggest where to get started? It looks like `test_ui.py` is the right place, but I see in there colors are explicitly turned off. Would it be possible / a good idea to turn them on?

PR #805, opened 2018-04-16
fixed powershell coloring Bug #772

afwilkin:
The issue was that running theFuck would change the color of the powershell. This was more of a powershell issue than an issue of theFuck. However, by running "[Console]::ResetColor()" at the end of the powershell command to run theFuck we were able to remedy the issue. #772


nvbn:
Thanks!

PR #810, opened 2018-05-07
Add new `long_form_help` rule

jakewarren:
Adds a new rule to help deal with programs that don't understand the `-h` short form of `--help`. Also attempts to determine the proper help command suggested by the tool (such as with `go build -h` -> `go help build`). 

I don't work with Python much and this is my first time using pytest so I am open to suggestions.


nvbn:
Thanks!

Changed code a bit for better suiting our code style and squashed it manually.

PR #812, opened 2018-05-09
Update README.md to include chromeos install

Ewpratten:


Ewpratten:
appveyor seems to be having problems with pip..

scorphus:
Nice! Thanks, @Ewpratten.

ps.: the AppVeyor build fix is on its way.

PR #816, opened 2018-05-22
Add brew_reinstall rule

revolter:
Replaces install with reinstall when a package is already installed.


revolter:
Wow, thanks for the quick merge!

revolter:
Inception:

screen shot 2018-05-23 at 11 38 31

PR #818, opened 2018-06-09
git_push: Handle command containing force argument

gma2th:
```
git push --force
fatal: The current branch [branch_name] has no upstream branch.

    git push --set-upstream origin [branch_name]
```

Before:
`git push --force --force`
After:
`git push --set-upstream origin master --force`

command.output:
```
03:32:30.280527 git.c:344               trace: built-in: git push --force
fatal: The current branch [branch_name] has no upstream branch.

    git push --set-upstream origin [branch_name]
```
Regex stopped at the first `git push` with `--force` argument.
Not sure why the command is part of the output though?

### DIFF
--- a/tests/rules/test_git_push.py
+++ b/tests/rules/test_git_push.py
@@ -66,6 +66,10 @@ def test_not_match(output, script, branch_name):
     ('git -c test=test push --quiet origin', 'master',
      'git -c test=test push --set-upstream origin master --quiet'),
     ('git push', "test's",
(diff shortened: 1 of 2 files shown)


scorphus:
Hi @gma2th,

Thank you for contributing to TheFuck.

The tests added in this PR pass even without the change in `thefuck/rules/git_push.py`.

Can you please add tests that fail with the previous revision?

nvbn:
Thanks!

Issue #711, opened 2017-10-16
Command selection of  `ctrl+n` and `ctrl+p` instead of `↑` and `↓`


ikuwow:
Feature request.

It's convinient to select commands by `ctrl+n` and `ctrl+p` instead of `↑` and `↓`.

```
$ gitstatus
-bash: gitstatus: command not found
$ fuck
gstat [enter/↑/↓/ctrl+c] # <= here, selection can be changed by ctrl+p or ctrl+n
```

ikuwow:
It's released. Thank you @nvbn  :+1:

PR #819, opened 2018-06-12
Fix get_key on Windows

MattKotsenas:
PR #711 moved the arrow and cancel key codes to `const.py`. However, the move also changed the codes from byte arrays to strings, which broke the use of `msvcrt.getch()` for Windows. This resulted in PowerShell users being unable to cancel with `Ctrl+C` (pressing `q` still worked).

The fix is to use `msvcrt.getwch()` so the key is a Unicode character, matching the Unix implementation.

### DIFF
--- a/thefuck/system/win32.py
+++ b/thefuck/system/win32.py
@@ -1,5 +1,4 @@
 import os
-import sys
 import msvcrt
 import win_unicode_console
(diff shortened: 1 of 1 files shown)


nvbn:
Thanks!

Issue #718, opened 2017-10-26
Cannot run fuck in zsh without export errors


icithis:
Installed via pip. Ran eval "$(thefuck --alias)" in both bash and zsh. $ fuck then produces 'No fucks given' in bash while in zsh it complains about invalid arguments.
```
$ eval "$(thefuck --alias)"
$ fuck
fuck:export:3: not valid in this context: -'
fuck:3: not an identifier: [skip
```

Also FWIW:
```
$ zsh --version
zsh 4.3.11 (x86_64-redhat-linux-gnu)
$ cat /etc/redhat-release
Red Hat Enterprise Linux Server release 6.8 (Santiago)
```

iraquitan:
Almost same problem here:
```zsh
$ fuck
fuck:export: not valid in this context: -'
pyenv:11: command not found: pyenv                                                                                                                                                                                                                                                                                                                                        
prompt_context:10: command not found: whoami
prompt_dir:1: command not found: sed
prompt_background_jobs:1: command not found: wc
prompt_background_jobs:2: command not found: awk
```

Additional info:
```zsh
$ zsh --version                                                                                                                                                                                                                                                                                                     
zsh 5.4.2 (x86_64-apple-darwin15.6.0)
$ sw_vers                                                                                                                                                                                                                                                                                                           
ProductName:	Mac OS X
ProductVersion:	10.11.6
BuildVersion:	15G17023

```

yeonzi:
it seems that was a issue of zsh.

when run `export TF_SHELL_ALIASES=$(alias)` in zsh, you can see the same error.

run `unalias -- -` and this error will disappear, but another error comes.

my way to deal with this problem is not use `eval "$(thefuck --alias)"` in ~/.zshrc

run `thefuck --alias` in your shell, delete the line which contain `export TF_SHELL_ALIASES=$(alias)`, and write the output in your .zshrc

just like

```
fuck () {
    TF_PYTHONIOENCODING=$PYTHONIOENCODING;
    export TF_ALIAS=fuck;
    export TF_HISTORY="$(fc -ln -10)";
    export PYTHONIOENCODING=utf-8;
    TF_CMD=$(
        thefuck THEFUCK_ARGUMENT_PLACEHOLDER $@
    ) && eval $TF_CMD;
    unset TF_HISTORY;
    export PYTHONIOENCODING=$TF_PYTHONIOENCODING;
    test -n "$TF_CMD" && print -s $TF_CMD
}
```

[...]

PR #820, opened 2018-06-15
Avoid masking shell return values in Zsh.app_alias

chrismendis:
I discovered that a common shell script issue (https://github.com/koalaman/shellcheck/wiki/SC2155) is present in the script returned by `Zsh.app_alias()`, amongst other `app_alias` methods in thefuck.

In the case of the zsh `app_alias()` script, this common issue is causing the script to error on some versions of Mac OS X. This is  reported in #718.

Unmasking the value of `TF_SHELL_ALIASES` fixes the errors in Mac OS X, but I also unmasked `TF_HISTORY` for consistency.


nvbn:
Thanks!

PR #822, opened 2018-07-05
added --yeah as an alternative arg to -y and --yes; updated README.md

Neutrollized:
Love this little app -- I use it a little too often...

In any case, I just thought you missed a good opportunity to do `fuck --yeah`, so I'm adding it in :)


nvbn:
Looks fun, thanks!

PR #823, opened 2018-07-07
Create `cat_dir` rule for replacing `cat` with `ls`

scolby33:
Examples

```sh
$ cat /etc
cat: /etc: Is a directory
$ fuck
ls /etc  [enter/↑/↓/ctrl+c]
*snip*
$ cat tests
cat: tests: Is a directory
$ fuck
ls tests [enter/↑/↓/ctrl+c]
*snip*
```

### DIFF
--- a/tests/rules/test_cat_dir.py
+++ b/tests/rules/test_cat_dir.py
@@ -1,0 +1,28 @@
+import pytest
+from thefuck.rules.cat_dir import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('command', [
+    Command('cat foo', 'cat: foo: Is a directory'),
+    Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory'),
+])
+def test_match(command):
+    assert match(command)
+
+
+@pytest.mark.parametrize('command', [
+    Command('cat foo', 'foo bar baz'),
+    Command('cat foo bar', 'foo bar baz'),
+    Command('', ''),
+])
+def test_not_match(command):
+    assert not match(command)
+
+
+@pytest.mark.parametrize('command, new_command', [
+    (Command('cat foo', 'cat: foo: Is a directory'), 'ls foo'),
+    (Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory'), 'ls /foo/bar/'),
+])
+def test_get_new_command(command, new_command):
+    assert get_new_command(command) == new_command
--- a/tests/rules/test_cat_dir.py
+++ b/tests/rules/test_cat_dir.py
@@ -4,8 +4,9 @@ from thefuck.types import Command
 
 
 @pytest.mark.parametrize('command', [
-    Command('cat foo', 'cat: foo: Is a directory'),
-    Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory'),
+    Command('cat foo', 'cat: foo: Is a directory\n'),
+    Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory\n'),
+    Command('cat cat/', 'cat: cat/: Is a directory\n'),
 ])
 def test_match(command):
     assert match(command)
@@ -14,15 +15,15 @@ def test_match(command):
 @pytest.mark.parametrize('command', [
     Command('cat foo', 'foo bar baz'),
     Command('cat foo bar', 'foo bar baz'),
-    Command('', ''),
 ])
 def test_not_match(command):
     assert not match(command)
 
 
 @pytest.mark.parametrize('command, new_command', [
-    (Command('cat foo', 'cat: foo: Is a directory'), 'ls foo'),
-    (Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory'), 'ls /foo/bar/'),
+    (Command('cat foo', 'cat: foo: Is a directory\n'), 'ls foo'),
+    (Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory\n'), 'ls /foo/bar/'),
+    (Command('cat cat', 'cat: cat: Is a directory\n'), 'ls cat'),
 ])
 def test_get_new_command(command, new_command):
     assert get_new_command(command) == new_command
--- a/thefuck/rules/cat_dir.py
+++ b/thefuck/rules/cat_dir.py
@@ -1,0 +1,12 @@
+import re
+
+
+def match(command):
+    return (
+        command.script.startswith('cat') and
+        re.match(r'cat: .+: Is a directory', command.output)
+    )
+
+
+def get_new_command(command):
+    return re.sub(r'^cat', 'ls', command.script)
--- a/thefuck/rules/cat_dir.py
+++ b/thefuck/rules/cat_dir.py
@@ -1,12 +1,10 @@
-import re
-
-
 def match(command):
     return (
         command.script.startswith('cat') and
-        re.match(r'cat: .+: Is a directory', command.output)
+        command.output.startswith('cat: ') and
+        command.output.rstrip().endswith(': Is a directory')
     )
 
 
 def get_new_command(command):
-    return re.sub(r'^cat', 'ls', command.script)
+    return command.script.replace('cat', 'ls', 1)
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ following rules are enabled by default:
 * `aws_cli` &ndash; fixes misspelled commands like `aws dynamdb scan`;
 * `cargo` &ndash; runs `cargo build` instead of `cargo`;
 * `cargo_no_command` &ndash; fixes wrongs commands like `cargo buid`;
+* `cat_dir` &ndash; replaces `cat` with `ls` when you try to `cat` a directory
 * `cd_correction` &ndash; spellchecks and correct failed cd commands;
 * `cd_mkdir` &ndash; creates directories before cd'ing into them;
 * `cd_parent` &ndash; changes `cd..` to `cd ..`;


scorphus:
Hi @scolby33, thanks for contributing to TheFuck.

In Python, we should always try to solve the problem with [string methods](https://docs.python.org/3/library/stdtypes.html#string-methods) – such as `endswith`, `startswith`, `lstrip`, `rstrip`, etc. – before resorting to regular expressions, which are slower most of times.

So, I'll add a few line comments in order to suggest some changes to your PR regarding the above.

tests/rules/test_cat_dir.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+from thefuck.rules.cat_dir import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('command', [
+    Command('cat foo', 'cat: foo: Is a directory'),
+    Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory'),
```
scorphus: The actual output contains a LINEFEED (`\n`) at the end, so should the test output: `'cat: foo: Is a directory\n'` and `'cat: /foo/bar/: Is a directory\n'`

thefuck/rules/cat_dir.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+
+def match(command):
+    return (
+        command.script.startswith('cat') and
+        re.match(r'cat: .+: Is a directory', command.output)
```
scorphus: `re.match` could be replaced with `str.startswith`, `str.rstrip` and `str.endswith` used in conjunction. `str.rstrip` is needed here because `command.output` ends with `\n` (see previous line comment)

thefuck/rules/cat_dir.py
```diff
@@ -0,0 +1,12 @@
...
+    )
+
+
+def get_new_command(command):
+    return re.sub(r'^cat', 'ls', command.script)
```
scorphus: This can be easily replaced with [`str.replace`](https://docs.python.org/3/library/stdtypes.html?highlight=replace#str.replace). Just make sure that only the first occurrence of `cat` is replaced.

tests/rules/test_cat_dir.py
```diff
@@ -0,0 +1,28 @@
...
+])
+def test_match(command):
+    assert match(command)
+
+
+@pytest.mark.parametrize('command', [
+    Command('cat foo', 'foo bar baz'),
+    Command('cat foo bar', 'foo bar baz'),
+    Command('', ''),
```
scorphus: This test parameter can be removed.

thefuck/rules/cat_dir.py
scolby33: I wouldn't have considered the case of more than one `cat`. I added a test for this as well.

thefuck/rules/cat_dir.py
scolby33: What are the relative merits of `command.output.rstrip().endswith(': Is a directory')` vs. `command.output.endswith(': Is a directory\n')`?

scolby33:
Thanks for the feedback! I've made the changes you requested and I'll push once tox finishes running. Let me know if you see anything else.

I've tested the output of `cat` on macOS, Debian, and Arch. If someone out there is running a more esoteric system, it'd be interesting to check if all `cat`'s respond the same way.

thefuck/rules/cat_dir.py
scorphus: Well, as you probably guessed, my initial thought was that there may be a version – or a user defined alias or function – of `cat` that doesn't add that trailing newline. Then I went on and did some benchmarks and it turns out `rstrip` makes it around 20% slower – not that much. Maybe it's worth having.

Or maybe we're better off with just `... endswith(': Is a directory\n')` and those who use a `cat` that doesn't follow “default” implementation are ”in a state of sin and deserve whatever happens to them.” We can still have ` ... or endswith(': Is a directory')` if we ever need to, after all.

I wonder what your thoughts on this are.

scolby33: Here's some of my own benchmarks
```python
# bench.py
import timeit

# the base case
rstrip_timer = timeit.Timer(r"'cat: foo: Is a directory\n'.rstrip().endswith(': Is a directory')")
# what if there's no newline?
rstrip_timer2 = timeit.Timer(r"'cat: foo: Is a directory'.rstrip().endswith(': Is a directory')")
# what if it's a mess?
rstrip_timer3 = timeit.Timer(r"'cat: foo: Is a directory \n\n \n'.rstrip().endswith(': Is a directory')")
# base case assuming one newline
no_rstrip_timer = timeit.Timer(r"'cat: foo: Is a directory\n'.endswith(': Is a directory\n')")

timers = [rstrip_timer, rstrip_timer2, rstrip_timer3, no_rstrip_timer]

try:
    number = max(timer.autorange()[0] for timer in timers)
except AttributeError:  # no autorange in Python 2
    number = 2000000
print('number = {:,}'.format(number))

results = [min(timer.repeat(repeat=5, number=number)) for timer in timers]
print('results = {}'.format(results))
results_percentages = ['{:.2%}'.format((result-results[0])/results[0]) for result in results]
print('percentages = {}'.format(results_percentages))  # bigger negative is better
```

```fish
fish $ for version in 3.7 3.6 2.7
           echo "Python $version"
           eval python$version bench.py
       end
Python 3.7
number = 2,000,000
results = [0.46968697199999987, 0.41365929999999995, 0.48141546300000027, 0.3026672379999997]
percentages = ['0.00%', '-11.93%', '2.50%', '-35.56%']
Python 3.6
number = 10,000,000
results = [2.472023139009252, 2.1997297610214446, 2.5396445730002597, 1.6017803400172852]
percentages = ['0.00%', '-11.02%', '2.74%', '-35.20%']
Python 2.7
number = 2,000,000
results = [0.48915696144104004, 0.4453470706939697, 0.563338041305542, 0.329495906829834]
percentages = ['0.00%', '-8.96%', '15.17%', '-32.64%']
```

It's interesting that Python 3.6 picks such a bigger number in autorange, but the relative timing differences are approximately the same in all Python versions. Also interesting is that the case with all kinds of screwed up whitespace is *super* slow in Python 2.7.

I think that 35% lower times are a win, since the match code is (at least in my guess) the most performance-sensitive part of thefuck.

I think using a custom alias/wrapper is both "sinful" enough and unlikely enough to not be worth supporting in the face of this performance difference. I would thus vote for the non-`rstrip` version unless anyone can find a default implementation of `cat` that outputs different whitespace or someone comes back in the future complaining that the rule doesn't work for their `cat`.

scorphus: WOW! Super! This is how far we go for open source! 🎉

Let me share my benchmarks with you:

Name (time in ns)    |        Min          |        Max          |       Mean          | StdDev          |     Median          |    IQR          | Outliers  OPS (Kops/s)          | Rounds | Iterations
---------------------|---------------------|---------------------|---------------------|-----------------|---------------------|-----------------|---------------------------------|--------|-----------
test_with_endswith   |   422.3146 (1.00)   |   441.1720 (1.02)   |   428.6322 (1.01)   | 7.4089 (4.96)   |   426.0483 (1.00)   | 7.9161 (4.70)   |      1;0    2,333.0027 (0.99)   |      5 |   10000000
test_with_rstrip     |   528.8423 (1.25)   |   532.5924 (1.24)   |   531.3621 (1.25)   | 1.4928 (1.0)    |   531.7795 (1.25)   | 1.6846 (1.0)    |      1;0    1,881.9557 (0.80)   |      5 |    1895592
test_with_match      | 1,164.0312 (2.76)   | 1,171.8660 (2.72)   | 1,167.8383 (2.74)   | 3.7479 (2.51)   | 1,166.3237 (2.75)   | 6.9720 (4.14)   |      3;0      856.2829 (0.36)   |      5 |    1000000

Generated from:

```python
import pytest
import re


class TestEndswithRstripMatch(object):

    text = 'cat: /foo/bar/: Is a directory\n'

    def with_match(self):
        return re.match(r'cat: .+: Is a directory', self.text)

    @pytest.mark.benchmark(group='endswith vs rstrip vs match')
    def test_with_match(self, benchmark):
        result = benchmark(self.with_match)
        assert result

    def with_rstrip(self):
        return self.text.startswith('cat: ') and self.text.rstrip().endswith(': Is a directory')

    @pytest.mark.benchmark(group='endswith vs rstrip vs match')
    def test_with_rstrip(self, benchmark):
        result = benchmark(self.with_rstrip)
        assert result

    def with_endswith(self):
        return self.text.startswith('cat: ') and self.text.endswith(': Is a directory\n')

    @pytest.mark.benchmark(group='endswith vs rstrip vs match')
    def test_with_endswith(self, benchmark):
        result = benchmark(self.with_endswith)
        assert result
```

You might as well need `$ pip install pytest-benchmark` and [these docs](http://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options).

So, sure. Let's go for the `rstrip`less version and make it more tolerant in the future, if necessary.

scolby33: Sounds good! I'm happy with my changes, so unless there's anything else you or other collaborators see needing to be done, I think I'm ready for the merge. Let me know if you would like the commits squashed.

nvbn:
Thanks!

Rom1deTroyes:
> Or maybe we're better off with just `... endswith(': Is a
> directory\n')` and those who use a `cat` that doesn't follow “default”
> implementation are ”in a state of sin and deserve whatever happens to
> them.” We can still have ` ... or endswith(': Is a directory')` if we
> ever need to, after all.

My sin is on an other kind of state :

	romain@dulix:~/Musique/White_Cowbell_Oklahoma$ cat ..
	cat: ..: est un dossier
	

One more time, I miss a test (-:

> I wonder what your thoughts on this are.

++
    Romain
-- 
Ceux qui ne savent pas où ils vont sont surpris d'arriver ailleurs.
	-+- Pierre Dac -+-

scolby33:
I’m not super familiar with internationalization, but a bit of digging brought me to `gettext`. The closest I can find to the string we’re seeing in English is “Is a directory” (note the leading capital) in glibc’s `sysdeps/gnu/errlist.c`, which is translated to “est un dossier” on [the translation project](https://translationproject.org/domain/libc.html). I suppose we could parse out the messages from the files there or perhaps the python `gettext` library is what we want.

scolby33:
I’m on mobile so can’t try it right now, and I probably should be asleep, but from a quick look at the docs, it might be as simple as `gettext.translation(domain=‘libc’).gettext(‘Is a directory’)`

scolby33:
@Rom1deTroyes could you post the output of `locale` and the values of the environment variables `LANGAUGE`, `LC_ALL`, `LC_MESSAGES`, and `LANG` from your system?

scolby33:
I've opened #826 to add localization to this rule.

nvbn:
@Rom1deTroyes are you using experimental instant mode or shell logger? Otherwise thefuck should rerun the command with `LANG=C` and get error message in English.

Rom1deTroyes:
Le mardi 10 juil. 2018 à 11:40:38 (-0700), Scott Colby a écrit :
> @Rom1deTroyes could you post the output of `locale` and the values of
> the environment variables `LANGAUGE`, `LC_ALL`, `LC_MESSAGES`, and
> `LANG` from your system?

$ locale
LANG=fr_FR.UTF-8
LANGUAGE=
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL=


BTW, there is no C or C.utf8 locales on this system (I don't know why :
this is an old eepc used as a personal assistant) :

$ locale -a
fr_FR.utf8
POSIX

++
    Romain
-- 
Ce qu'on fait n'est jamais compris mais seulement loué ou blâmé.
	-+- Friedrich Nietzsche, Le gai savoir -+-

PR #826, opened 2018-07-10
Add cat dir localization

scolby33:
See https://github.com/nvbn/thefuck/pull/823#issuecomment-403783966 and following comments.

### DIFF
--- a/tests/rules/test_cat_dir.py
+++ b/tests/rules/test_cat_dir.py
@@ -1,15 +1,63 @@
+import gettext
+import os
+import warnings
+
 import pytest
+
 from thefuck.rules.cat_dir import match, get_new_command
 from thefuck.types import Command
 
+# patch this so the libc locale is available in the testing virtualenv
+gettext._default_localedir = '/usr/share/locale'
 
-@pytest.mark.parametrize('command', [
-    Command('cat foo', 'cat: foo: Is a directory\n'),
-    Command('cat /foo/bar/', 'cat: /foo/bar/: Is a directory\n'),
-    Command('cat cat/', 'cat: cat/: Is a directory\n'),
-])
-def test_match(command):
-    assert match(command)
+lang_vars = ['LANGUAGE', 'LC_ALL', 'LC_MESSAGES', 'LANG']
+
+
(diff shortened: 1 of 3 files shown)


tests/rules/test_cat_dir.py
```diff
@@ -1,15 +1,63 @@
...
+import os
+import warnings
+
 import pytest
+
 from thefuck.rules.cat_dir import match, get_new_command
 from thefuck.types import Command
 
+# patch this so the libc locale is available in the testing virtualenv
+gettext._default_localedir = '/usr/share/locale'
```
scolby33: This is dirty and I don't like it, but the default `_default_localedir` includes `sys.base_prefix` and `share/locale` isn't generally linked into a virtual environment.

nvbn:
At the moment by default thefuck gets output by reruning command with `LANG=C`, so the output should be in English. Only in the experimental instant mode or with shell logger thefuck may use translated output.

As we will need to support translations in other rules in the future, I think we will need something like this but more generic.

But for this rule, maybe it will be better to just check that the command is `cat` and an argument is a directory?

scolby33:
Rerunning with `LANG=C` is super smart! I wonder what the original commenter's configuration is that that didn't work.

`gettext` seems like a moderately good system for adapting to translations of other commands, but I think you'll find there to be some difficulties that will make it hard to keep it completely generic, unfortunately--it should be a ~hard~fun project.

As for modifying this rule, I think something like `os.path.isdir(command.script[4:])` could do it?

scolby33:
@nvbn see my edit above, would using pytest to create a temporary directory make sense or is that going too far?

nvbn:
A bit too far =)

I think it will be enough to just mock, like in [prove_recursively rule](https://github.com/nvbn/thefuck/blob/985b2d9ec9d23a16d00a8288024722207189326a/tests/rules/test_prove_recursively.py#L10).

scolby33:
I'm a bit sad that my beautiful 100-test matrix won't get to see the light of day, but I agree that this approach is much more elegant. I've opened a new pull request so I didn't have to fight with git to revert my initial set of changes.

See #827.

PR #830, opened 2018-07-18
#367: Support BSD style output in touch rule

scorphus:
On a Mac, also on NetBSD or OpenBSD, `touch` errs differently:

```
$ uname; touch a/b/c
Darwin
touch: a/b/c: No such file or directory
```

That gets matched by the rule but not fixed by it. Thus the regex pattern is now a bit more tolerant.

Thanks for any review and/or comment


nvbn:
Thanks!

PR #834, opened 2018-08-14
Added a rule to match az binary sub-command misses

jamtur01:


jamtur01:
Heavily lifted from the `aws_cli.py` rule.

jamtur01:
I'll add a test. Missed the individual rule tests.

PR #844, opened 2018-10-01
Add the pwsh support for thefuck

### LINKED ISSUE #832: thefuck does not support pwsh
>The output of `thefuck --version` (something like `The Fuck 3.1 using Python 3.5.0`):

* The Fuck 3.27 using Python 3.7.0

>Your shell and its version (`bash`, `zsh`, *Windows PowerShell*, etc.):

 * PowerShell Core 6.0.2

>Your system (Debian 7, ArchLinux, Windows, etc.):

 * Windows 10.0.7134 (April Creator's Update)

>How to reproduce the bug:

Attempt to invoke thefuck while using PowerShell Core (aka `pwsh`)

>The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

Skipping this one. Will explain why below.

>If the bug only appears with a specific application, the output of that application and its version:

N/A

>Anything else you think is relevant:

So the issue is actually in [thefuck's main `__init.py__`](https://github.com/nvbn/thefuck/blob/master/thefuck/shells/__init__.py). The mapping between shell process name and shell type is incomplete. PowerShell Core's process name is `pwsh`. This doesn't exist, so when running `pwsh`, thefuck just uses __init__'s `Generic()`, which seems to treat the environment as a bash-alike.

I'm pretty sure the fix is simple: change the mapping to look like this:

```python
shells = {'bash': Bash,
          'fish': Fish,
          'zsh': Zsh,
          'csh': Tcsh,
          'tcsh': Tcsh,
          'powershell': Powershell,
          'pwsh': Powershell}
``` 

I can open a PR for this, but I don't think I have the bandwidth to fully set up the environment or do testing on this, so I'd rather leave this breadcrumb for someone else to pick up on.

manikantanallagatla:
Update: attempt to fix #832.

### DIFF
--- a/thefuck/shells/__init__.py
+++ b/thefuck/shells/__init__.py
@@ -16,7 +16,8 @@ shells = {'bash': Bash,
           'zsh': Zsh,
           'csh': Tcsh,
           'tcsh': Tcsh,
-          'powershell': Powershell}
+          'powershell': Powershell,
+          'pwsh': Powershell}
 
 
 def _get_shell_from_env():


scorphus:
Thanks for your contribution, @manikantanallagatla! Did you test it with PowerShell on Windows?

nvbn:
Thanks!

PR #846, opened 2018-10-02
Some improvements

scorphus:
**Please don't merge yet**. I'd like these changes to be reviewed. Once it's all good I'll push to master. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,7 +390,8 @@ Several *The Fuck* parameters can be changed in the file `$XDG_CONFIG_HOME/thefu
 * `history_limit` &ndash; numeric value of how many history commands will be scanned, like `2000`;
 * `alter_history` &ndash; push fixed command to history, by default `True`;
 * `wait_slow_command` &ndash; max amount of time in seconds for getting previous command output if it in `slow_commands` list;
-* `slow_commands` &ndash; list of slow commands.
+* `slow_commands` &ndash; list of slow commands;
+* `num_close_matches` &ndash; maximum number of close matches to suggest, by default `3`.
 
 An example of `settings.py`:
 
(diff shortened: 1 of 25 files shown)


README.md
```diff
@@ -420,7 +422,8 @@ rule with lower `priority` will be matched first;
 * `THEFUCK_HISTORY_LIMIT` &ndash; how many history commands will be scanned, like `2000`;
 * `THEFUCK_ALTER_HISTORY` &ndash; push fixed command to history `true/false`;
 * `THEFUCK_WAIT_SLOW_COMMAND` &ndash; max amount of time in seconds for getting previous command output if it in `slow_commands` list;
-* `THEFUCK_SLOW_COMMANDS` &ndash; list of slow commands, like `lein:gradle`.
+* `THEFUCK_SLOW_COMMANDS` &ndash; list of slow commands, like `lein:gradle`;
+* `THEFUCK_NUM_CLOSE_MATCHES` &ndash; maximum number of close matches to suggest, like `5`.
```
josephfrazier: Should we just switch to putting a `;` on the last line (instead of `.`), so the diff can be smaller going forward?

scorphus:
Thanks for the review, @josephfrazier! "Rebase and merge", absolutely! I also prefer it :-)

README.md
scorphus: I had the exact same thought. But the next thought was "Wait. It's text, not code". Perhaps it makes sense to have a period there. For me it will look like an unfinished list.

scorphus: I believe that's a good one, reasonable change. :+1:

josephfrazier: Thanks, I'll wait until this is closed to make that change (if I remember)

scorphus:
Oh, let me add some tests to `thefuck/output_readers/rerun.py`.

README.md
nvbn: @josephfrazier I think it should be ok to remove `;`, as it's not required by grammar rules

nvbn:
Diff is huge, thanks 😀!

PR #848, opened 2018-10-11
Update README.md

khaledosman:
rename the --yeah argument to --yes
if you run fuck --help it doesn't output a --yeah flag, I believe it has been renamed to --yes
``` fuck --yeah ``` sounded cooler though


steaksauce-:
It seems like the option for --yeah is in the master branch, but is missing from the version releases.

PR #851, opened 2018-10-21
Added Persian letters rule for converting those letters to English

hossainalhaidari:
Sometimes when typing in shell, we're in a wrong keyboard layout (like Persian, in this particular rule). What happens is, the entered command will be something like this:

`لهف سفشفعس`

In which if the keyboard layout was correct (i.e. English), it would be:

`git status`

This rule will fix these mistakes.

### DIFF
--- a/tests/rules/test_persian_letters.py
+++ b/tests/rules/test_persian_letters.py
@@ -1,0 +1,21 @@
+import pytest
+from thefuck.types import Command
+from thefuck.rules.chmod_x import match, get_new_command
+
+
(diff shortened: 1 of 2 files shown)


nvbn:
Thanks, but there's already [switch_lang](https://github.com/nvbn/thefuck/blob/master/thefuck/rules/switch_lang.py) rule. I think it will be better to modify it instead of creating a new rule.

hossainalhaidari:
Ah sorry I didn't notice that!

PR #853, opened 2018-10-25
Comply to new flake8 3.6

scorphus:
With new flake8 version 3.6 comes new errors, `W504` and `W605`. They didn't exist before the new version. `W504` is ignored along with `W503`. Fixes regarding `W605` are included.

Remember not to squash the commits ‒ for some good reasons. :-) Just rebase them.


nvbn:
Thanks!

PR #854, opened 2018-10-30
Remove zsh instant mode log with -f

longsleep:
In some setups, rm might default to interactive promt. This change adds
the -f parameter to force remove the instant mode log on exit to avoid
an interactive prompt.

```
 ~ 
rm: remove regular file
'/tmp/user/1000/thefuck-script-log-bbb81260140c4b3fa18bf2097f15bd77'?
```


nvbn:
Thanks!

Issue #855, opened 2018-11-09
Layout switching for Ukrainian is not supported, the Russian one is used instead


kozar:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.25 using Python 3.4.9

Your system (Debian 7, ArchLinux, Windows, etc.):

    Oracle Linux Server release 7.5

How to reproduce the bug:

execute command consisting of two or more parts using Ukrainian keyboard layout, e.g. (it's `git st`)
   
    $ пше іе

The fuck will suggest the

    git іt 

Instead of 

    git st

Anything else you think is relevant:

    Matching rule doesn't take into account all the command line, so it assumes the Russian layout is a suitable one for the replacement, even though it doesn't contain all needed characters

PR #856, opened 2018-11-09
#855 - Support Ukrainian layout; Fix matching of similar layouts

kozar:
Matching rule doesn't take into account all the command line, so if the first part of the command line contains only characters from the layout then it selected as suitable event if the rest of the line contains characters from other layouts. It's relevant for any similar layouts.
Also, the Ukrainian layout is added to the list of supported ones.


nvbn:
Thanks!

PR #861, opened 2018-12-08
Fix invalid for_app/sudo_support ordering

chrisdepas:
- Fixed incorrect ordering of for_app/sudo_support annotations causing apt_invalid_operation and dnf_no_such_command rules to fail
- The previous ordering meant that `for_app` was on the inside (i.e. called first), so it would fail for commands executed with `sudo`


nvbn:
Thanks!

Issue #862, opened 2018-12-09
Greek keyboard mapping incorrect


GKFX:
(Ref #776)

The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):
```
~ $ shit --version
The Fuck 3.28 using Python 3.6.7 and Bash 4.4.23(1)-release
```

Your system (Debian 7, ArchLinux, Windows, etc.):

    Debian Buster

How to reproduce the bug:
```
~ $ ωιμ   # expected vim
bash: ωιμ: command not found
127|~ $ shit
cin [enter/↑/↓/ctrl+c]
Aborted
```

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

```
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': False,
 'no_colors': False,
 'num_close_matches': 3,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': PosixPath('/home/bateman/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
DEBUG: Received output: /bin/sh: 1: ωιμ: not found

[...]

PR #864, opened 2018-12-13
Fix issue #862

RealOgre:
I create a dictionary that correlates a  Greek character to the appropriate English character. Some of the symbols are not in the dictionary like the " " and "-" but because this two characters are at the same position between the keyboards so no correlation is needed. 
The code had no errors when i run py.test.
I can provide more information or extra work, if deems necessary.


RealOrge:
If you like the solution with the dictionary i can do the same with Russian language

nvbn:
Thanks!

I've changed it a bit to make it possible to use similar solution for other languages in the future.

Merged manually in https://github.com/nvbn/thefuck/commit/8093f7cab800089ee1007c4b9e3e3e5479fcac3d

PR #873, opened 2019-01-05
Use `yield from`

abdulniyaspm:
Since the requirement is `python (3.4+)` we could use the shortened form of `for item in iterable`. ie, This PR suggest to change

```
 for command in sorted_commands:
       yield command
```
with

```
yield from sorted_commands
```
This is available from `python 3.3` onwards

### DIFF
--- a/thefuck/corrector.py
+++ b/thefuck/corrector.py
@@ -74,8 +74,7 @@ def organize_commands(corrected_commands):
     logs.debug('Corrected commands: '.format(
         ', '.join(u'{}'.format(cmd) for cmd in [first_command] + sorted_commands)))
 
-    for command in sorted_commands:
-        yield command
+    yield from sorted_commands
 
 
 def get_corrected_commands(command):


scorphus:
Thanks for your PR, @abdulniyaspm.

While I'd like to approve and accept this change – as much as many others – I believe we can and should [wait](https://pythonclock.org/) until Python 2 is eventually [discontinued](https://www.python.org/dev/peps/pep-0373/#update). I think it's fair to all users.

scorphus:
I'd say it still is and will probably be for some time.

Once Python 2 is officially EOL'ed, we can then release a new version that is still compatible with both 2 and 3 and set it as the last Py2-compatible one. Then, for the next releases, we state that they are Py3-compatible only – although they might still work on Py2.

And I think, moving forward, it will probably be easier implementing *new* features as we won't need to worry about Py2 anymore.

But, when I think about refactoring code and removing Py2-compatible blocks, I wonder if it's really worth the time and effort of doing so only for the sake of removing. What are the real advantages of that? Yes, a cleaner codebase, fewer test cases, a bit easier to maintain, etc. But is it really **that** much of difference? Is it really worth the time/effort or should we spend that time with something else?

In reality, we actually don't need to do anything when it comes to Py2 being discontinued. TheFuck will continue to work as Python evolves. It is and will be compatible with Python 3, as opposed to some other apps that aren't – [and won't ever be, it seems](https://bugs.launchpad.net/calibre/+bug/1714107) – compatible with Python 3.

djh82:
Python 2 (as well as 3.5) are now EoL.

scorphus:
We've been slow at drafting the last 2.7-compatible release, haven't we? So, next release will be the one.

PR #895, opened 2019-03-28
Add rule for pip_install permission fix

IngaFeick:
This PR reacts to `pip install` permission issues by either adding `--user` to the command, or, if `--user` is already in the command, removing the `--user` flag and running the command with sudo.

### DIFF
--- a/thefuck/rules/pip_install.py
+++ b/thefuck/rules/pip_install.py
@@ -12,4 +12,4 @@ def get_new_command(command):
     if '--user' not in command.script:  # add --user (attempt 1)
         return command.script.replace(' install ', ' install --user ')
 
-    return 'sudo {}'.format(command.script.replace(' --user', '')) # since --user didn't fix things, let's try sudo (attempt 2)
+    return 'sudo {}'.format(command.script.replace(' --user', ''))  # since --user didn't fix things, let's try sudo (attempt 2)
--- a/thefuck/rules/pip_install.py
+++ b/thefuck/rules/pip_install.py
@@ -11,5 +11,5 @@ def match(command):
 def get_new_command(command):
     if "--user" not in command.script:  # add --user (attempt 1)
         return command.script.replace(" install ", " install --user ")
-    else: # since --user didn't fix things, let's try sudo (attempt 2)
-    	return 'sudo {}'.format(command.script.replace(" --user", ""))        
+    else:  # since --user didn't fix things, let's try sudo (attempt 2)
+        return 'sudo {}'.format(command.script.replace(" --user", ""))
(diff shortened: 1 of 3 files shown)


nvbn:
The rule looks good, thanks!

Can you please fix [linter warnings](https://travis-ci.org/nvbn/thefuck/jobs/512578895#L419)?
```
./thefuck/rules/pip_install.py:14:10: E261 at least two spaces before inline comment
./thefuck/rules/pip_install.py:15:5: W191 indentation contains tabs
./thefuck/rules/pip_install.py:15:5: E101 indentation contains mixed spaces and tabs
./thefuck/rules/pip_install.py:15:68: W291 trailing whitespace
```

scorphus:
Thanks for the contribution, @IngaFeick! I'd recommend spending some time configuring your editor to do things like strip trailing whitespaces, standardise indentation, lint the source code, etc.

IngaFeick:
@nvbn @scorphus sorry mates, I forgot to run flake before submitting 🤦🏻‍♂️

thefuck/rules/pip_install.py
```diff
@@ -0,0 +1,15 @@
...
+@sudo_support
+@for_app('pip')
+def match(command):
+    return ('pip install' in command.script and "Permission denied" in command.output)
+
+
+def get_new_command(command):
+    if "--user" not in command.script:  # add --user (attempt 1)
+        return command.script.replace(" install ", " install --user ")
+    else:  # since --user didn't fix things, let's try sudo (attempt 2)
```
scorphus: This else here is unnecessary and should be removed.

IngaFeick: Quotation has been changed as requested. 
Could you kindly explain why the `else` is unnecessary? The script is supposed to react differently, depending on whether or not the command already included the `--user` flag, in which case sudo might fix the issue.

scorphus: An else after a return is unnecessary and only increases indentation, decreasing readability.

This:
```python
if condition:
    return foo
else:
    return bar
```
Is equivalent to this:
```python
if condition:
    return foo
return bar
```

IngaFeick: 🤦🏻‍♀️ you're absolutely right of course. I thought you asked me to delete the entire `else` block including the 2nd return statement.

IngaFeick: Changed as requested

scorphus: Oh, pardon me for not being clear about that :relaxed: Glad you changed it! Thanks for your contribution.

nvbn:
Looks good, thanks!

IngaFeick:
Great, thanks!

hongshaoyang:
Hate to be a wet blanket, but I don't think running sudo with pip install is a good idea. Essentially, you're allowing arbitrary Python code from PyPI to be executed with system-wide admin privileges. 

See [this](https://stackoverflow.com/a/37104948/6619250) and [this](https://stackoverflow.com/a/21056000/6619250).

PR #907, opened 2019-04-26
Adding yay AUR manager to Arch Linux's commands

jesusdanielcf:
Adding yay AUR manager to Arch Linux's commands since yaourt is unmantained and has some security issues.

Since those issues with yaourt it's recommended to use another AUR manager instead and yay is being the most used alternative.


scorphus:
Why is `root` the the author of the commit?

jesusdanielcf:
Hi! I did it from VM where I was trying thefuck and I didn't have the global configuration configured in that VM

nvbn:
Looks good, thanks!

PR #913, opened 2019-05-09
Create shoaib.py

shoaibsultan:
The first contribution in an open source project thefuck please guide through comments

### DIFF
--- a/thefuck/rules/shoaib.py
+++ b/thefuck/rules/shoaib.py
@@ -1,0 +1,16 @@
+# Adds the missing space between the ls command and the option (i.e. -l)
+# when trying to ls with options.
+#
+# Does not really save chars, but it is a fun and my first contribution ever in open source project.
+#
+# Example:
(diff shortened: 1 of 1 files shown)


djh82:
This is very specific and I would be surprised if this wasn't already covered by 'missing_space_before_subcommand'?

Regardless, you are missing some unit tests!

scorphus:
That's exactly the case. The [`missing_space_before_subcommand`](https://github.com/nvbn/thefuck/blob/master/thefuck/rules/missing_space_before_subcommand.py) rule already covers that. Check the [tests for one example](https://github.com/nvbn/thefuck/blob/2a166a7dec3c872ef6be3870b50889500ca6a212/tests/rules/test_missing_space_before_subcommand.py#L27). There, you can also see how to write test for a rule. Hoping to see more contributions from you, @shoaibsultan.

Thanks, @djh82! 🙌 Awesome to read from you!

Issue #921, opened 2019-06-02
[Bug] Recursive shell nesting


FlyingWombat:
TheFuck seems to recurse its prompt for the event where a shell is launched, directly or indirectly, from TheFuck's prompt.

The Fuck 3.29 (tested on git master 59dc6cbf906d910af04ff54fc339d85f837d0fe2)
Arch Linux 5.1.5 (x86_64)
Python 3.7.3 
ZSH 5.7.1 (same happens on Bash 5.0.7)

### Steps to reproduce

1. misspell call to shell: `bash` => `bsh` or `zsh` => `zh`
2. `fuck` 
3. Accept prompt to run `bash`
4. Now you should see TheFuck's prompt to run `bash` again.

### Log
The output of The Fuck with `THEFUCK_DEBUG=true`  
[thefuck-nested-shell.log](https://github.com/nvbn/thefuck/files/3244622/thefuck-nested-shell.log)

### Other Info

This doesn't just apply to launching a shell, but for other programs that launch shells as part of their operation. I discovered this because I use Ranger, a file manager where I often drop into a sub-shell. When I used TheFuck to launch Ranger, and then launch a sub-shell from Ranger, I encountered TheFuck's prompt again.

Testing was done on git-master, with TheFuck installed into a clean VirtualEnv.

scorphus:
Thanks for reporting, @FlyingWombat!

Can you please confirm de2e1d6 fixes the issue? [921-alias](/nvbn/thefuck/tree/921-alias) is the branch name. Thanks.

FlyingWombat:
Yes, the issue is fixed on `921-alias`

PR #923, opened 2019-06-04
#921: Try printing alias before trying to fix a command

scorphus:
The shell functions could be changed in a way to unset `TF_HISTORY` before firing the subprocess. But it would be prone to regressions. So, changing the order of `if` statements and adding a comment should suffice.


nvbn:
Looks good, thanks!

Issue #925, opened 2019-06-18
Help with `docker image rm`


Regaddi:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.29 using Python 3.7.3 and ZSH 5.7.1

### Your system (Debian 7, ArchLinux, Windows, etc.):

    macOS Mojave 10.14.5

How to reproduce the bug:

    Try to remove a docker image for which a container is still running.

```bash
$ docker image rm -f cd809b04b6ff
Error response from daemon: conflict: unable to delete cd809b04b6ff (cannot be forced) - image is being used by running container e5e2591040d1
$ fuck
docker image rm -f cd809b04b6ff [enter/↑/↓/ctrl+c]
```

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

<details>
<summary>Click for output</summary>

```bash
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': False,
 'no_colors': False,
 'num_close_matches': 3,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': PosixPath('/Users/me/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
DEBUG: Received output: Error response from daemon: conflict: unable to delete cd809b04b6ff (cannot be forced) - image is being used by running container e5e2591040d1

[...]

PR #928, opened 2019-07-05
Docker remove container before remove image

cjmartian:
This pr fixes https://github.com/nvbn/thefuck/issues/925

### DIFF
--- a/tests/rules/test_docker_image_being_used_by_container.py
+++ b/tests/rules/test_docker_image_being_used_by_container.py
@@ -1,0 +1,15 @@
+from thefuck.rules.docker_image_being_used_by_container import match, get_new_command
+from thefuck.types import Command
+
+
+def test_match():
+    err_response = """
+    Error response from daemon: conflict: unable to delete cd809b04b6ff (cannot be forced) - image 
+    is being used by running container e5e2591040d1 
+    """
+    assert match(Command('docker image rm -f cd809b04b6ff', err_response))
+
+
+def test_get_new_command():
+    assert get_new_command(Command('docker image rm -f cd809b04b6ff',
+                                   '')) == 'docker container rm -f e5e2591040d1 && docker image rm -f cd809b04b6ff'
--- a/tests/rules/test_docker_image_being_used_by_container.py
+++ b/tests/rules/test_docker_image_being_used_by_container.py
@@ -12,6 +12,11 @@ def test_not_match():
     assert not match(Command('docker image rm -f cd809b04b6ff', err_response))
 
 
+def test_not_docker_command():
+    err_response = """Error response from daemon: conflict: unable to delete cd809b04b6ff (cannot be forced) - image is being used by running container e5e2591040d1"""
+    assert not match(Command('git image rm -f cd809b04b6ff', err_response))
+
+
 def test_get_new_command():
     err_response = """
         Error response from daemon: conflict: unable to delete cd809b04b6ff (cannot be forced) - image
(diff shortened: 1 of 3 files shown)


thefuck/rules/docker_image_being_used_by_container.py
```diff
@@ -0,0 +1,13 @@
+from thefuck.utils import for_app
+
+
+@for_app('docker')
+def match(command):
+    return ('docker' in command.script
```
nvbn: `@for_app('docker')` already ensures that the `command.script` starts with `docker`.

thefuck/rules/docker_image_being_used_by_container.py
```diff
@@ -0,0 +1,13 @@
...
+            and 'image is being used by running container' in command.output)
+
+
+def get_new_command(command):
+    container_id = command.output.strip().split(' ')
+    print(container_id[-1])
```
nvbn: Better not to print here, it might be evaluated in some shells =)

thefuck/rules/docker_image_being_used_by_container.py
```diff
@@ -0,0 +1,13 @@
...
+    return 'docker container rm -f {} && {}'.format(container_id[-1], command.script)
```
nvbn: For compatibility with different shells it's better to use `shell.and_` from `thefuck.shells`.

tests/rules/test_docker_image_being_used_by_container.py
```diff
@@ -0,0 +1,17 @@
+from thefuck.rules.docker_image_being_used_by_container import match, get_new_command
+from thefuck.types import Command
+
+
+def test_match():
+    err_response = """Error response from daemon: conflict: unable to delete cd809b04b6ff (cannot be forced) - image is being used by running container e5e2591040d1"""
```
nvbn: Might be a good idea to also add test case for "not match".

thefuck/rules/docker_image_being_used_by_container.py
cjmartian: Oops! Thank you for catching that. I'll remove it!

cjmartian:
@nvbn I went ahead and made those changes, I hope this looks better now. Thank you for checking this out for me! :)

thefuck/rules/docker_image_being_used_by_container.py
```diff
@@ -0,0 +1,20 @@
...
+    return 'image is being used by running container' in command.output
+
+
+def get_new_command(command):
+    '''
+    Prepends docker container rm -f {container ID} to
+    the previous docker image rm {image ID} command
+    '''
+    container_id = command.output.strip().split(' ')
+    return shell.and_('docker container rm -f {} && {}').format(container_id[-1], command.script)
```
scorphus: The thing is, some shells – such as Fish<sup>1</sup> – do not support the `&&` operator. They use things like `; and` instead, for example. That's why each of the shells in TheFuck implement their own `AND` operator (i.e. `shell.and_()`. It takes a sequence of commands and joins them by the supported operator:

```python
oneliner = shell.and('cmd1', 'cmd2', 'cmd3')
```

So, please change it like the following:

```suggestion
    return shell.and_('docker container rm -f {}', '{}').format(container_id[-1], command.script)
```  
 
> * <sup>1</sup> Recent versions of Fish support `&&` but we must keep backward compatibility

cjmartian: That makes sense! I'll make that change. Thanks for the information.

nvbn:
Looks good, thanks!

cjmartian:
Thanks everyone! I'll be on the lookout for other issues I can fix. Feel free to tag me if there's something you want me to take on.

PR #931, opened 2019-07-12
Only consider raw command in output

tobixx:
... else it will not work for localized messages.

Example German output:
```
Führen Sie »apt list --upgradable« aus, um sie anzuzeigen.
```


tobixx:
@scorphus done

nvbn:
Looks good, thanks!

PR #933, opened 2019-07-19
Add: rules/go_unknown_command for misspelled go commands.

ik1ne:
- Add: rules/go_unknown_command for misspelled go commands.
- Add: tests/test_go_unknown_command which tests match and mismatch case of rules/go_unknown_command.
- Change: Added description of go_unknown_command to README.md.

### DIFF
--- a/thefuck/rules/go_unknown_command.py
+++ b/thefuck/rules/go_unknown_command.py
@@ -1,14 +1,13 @@
-from thefuck.utils import get_closest, replace_argument
+from thefuck.utils import get_closest, replace_argument, for_app
 
 _GOLANG_COMMANDS = (
     "bug", "build", "clean", "doc", "env", "fix", "fmt", "generate", "get",
     "install", "list", "mod", "run", "test", "tool", "version", "vet")
 
 
+@for_app('go')
 def match(command):
-    return (command.script_parts
-            and command.script_parts[0] == "go"
-            and 'unknown command' in command.output)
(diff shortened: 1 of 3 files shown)


ik1ne:
So obviously I missed the test for `get_new_command`, I'm going to add test for that and update the PR.

thefuck/rules/go_unknown_command.py
```diff
@@ -0,0 +1,18 @@
+from thefuck.utils import get_closest, replace_argument
+
+_GOLANG_COMMANDS = (
```
nvbn: Can you please make that list dynamic, like in https://github.com/nvbn/thefuck/blob/master/thefuck/rules/gem_unknown_command.py#L17 ?

thefuck/rules/go_unknown_command.py
```diff
@@ -0,0 +1,18 @@
...
+    "bug", "build", "clean", "doc", "env", "fix", "fmt", "generate", "get",
+    "install", "list", "mod", "run", "test", "tool", "version", "vet")
+
+
+def match(command):
+    return (command.script_parts
+            and command.script_parts[0] == "go"
```
nvbn: It will be better to replace that check with `for_app` decorator, it helps with things like `sudo`:
```python

@for_app('go')
def match(command):
    ...
```

ik1ne:
Done.

nvbn:
Awesome, thanks!

Issue #937, opened 2019-07-30
rules.docker_not_command works only in test environment.


ik1ne:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.29 using Python 3.7.4 and ZSH 5.3

Your system (Debian 7, ArchLinux, Windows, etc.):

    Mac OS X Mojave 10.14.6 (18G84)

How to reproduce the bug:

```shell
shell $ docker contner ls
shell $ thefuck
docker contner ls [enter/↑/↓/ctrl+c]
```

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

```shell
shell $ export THEFUCK_DEBUG=true
shell $ docker contner ls        
docker: 'contner' is not a docker command.
See 'docker --help'
shell $ fuck
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': False,
 'no_colors': False,
 'num_close_matches': 3,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': PosixPath('/Users/ik1ne/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
DEBUG: Received output: docker: 'contner' is not a docker command.
See 'docker --help'

[...]

PR #939, opened 2019-08-02
Docker update #937

ik1ne:
Newer version of docker outputs its result to stderr, not stdout.

Also, docker's command has been changed a lot (for example, `docker images` is now `docker image ls`).


ik1ne:
Sorry, I forgot to open this as draft.

PR #942, opened 2019-08-09
Rule for branch dash 0

ProfessorTom:
fixes fat-fingering `git branch -0` instead of `git branch -v`


nvbn:
Nice idea, maybe it will be a good thing to make that more generic? To just replace any argument that starts with `0` if it appears in `stderr`.

ProfessorTom:
@nvbn When I quickly read your comment on my phone, I thought making this fix more generic sounded like a good idea, but upon reflection, I'm not so sure.

In the specific case I coded up, the argument needs to turn into a flag. Would making that assumption more generic (i.e. convert all arguments starting with `0` to `-`) make sense?

For that matter, what would the "undo" action look like, or would there even need to be an "undo" action? (In the specific example here, I need to delete a branch that was just accidentally created...what should the proper response be if `git branch 0v` wasn't the command?

nvbn:
@ProfessorTom you're right, initially I thought that `git branch 0v` prints something, but it just creates a branch.

I guess the only way to generalize this rule is to also support cases like `git branch 0l`, `git branch 0a` and etc, so just check that the argument after `branch` starts with `0`.

ProfessorTom:
Do you still want me to generalize this feature turning the `0` into a `-` and deleting the branch just created in all cases?

Is there a case where this would cause more harm than good?

ProfessorTom:
bumping to get a review and hopefully a merge.

thefuck/rules/git_branch_0v_to_dash_v.py
```diff
@@ -0,0 +1,13 @@
+from thefuck.shells import shell
+from thefuck.specific.git import git_support
+
+
+@git_support
+def match(command):
+    return (command.script_parts
+            and command.script_parts[1:] == 'branch 0v'.split())
```
scorphus: No need to split. Could use a list instead.
```suggestion
            and command.script_parts[1:] == ['branch', '0v'])
```

Not that the way I use Git is of any importance, but I seldom use `-v`, if ever. I use `-r` or `-l` or `-d` much more frequently. So perhaps the matching should be more comprehensive. How about the following:

```python
if len(a) == 2 and a.startswith("0")
```

Then the rule could have a function to return the first occurrence of that in `script_parts`. Something like:

```python
    return next((a for a in script_parts if len(a) == 2 and a.startswith("0")), "")
```

That function could be [*memoized*](https://github.com/nvbn/thefuck/blob/master/thefuck/rules/git_add.py#L8) and used in `get_new_command` as well.

What do you think?

tests/rules/test_get_branch_0v_to_dash_v.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+from thefuck.rules.git_branch_0v_to_dash_v import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.fixture
+def output():
+    return ""
```
scorphus: This fixture can be removed and all it's occurrences be replaced by `""`.

tests/rules/test_get_branch_0v_to_dash_v.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_match(output):
+    assert match(Command('git branch 0v', output))
```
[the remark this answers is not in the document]
scorphus: No, unfortunately. That matter has been lurking in the back of my mind for some time now. I still need to discuss it with other maintainers.

So, that's more a suggestion of something that can be improved and not something mandatory. You hopefully noticed the "Nit:" prefix. 🙂 I try to do my best when reviewing and [Google Engineering's best practices](https://github.com/google/eng-practices) has helped in that regard. Although I now see I could have been more clear and/or linked the best practices under "Nit:" — as in [Nit](https://github.com/google/eng-practices/blob/bfe316ac6f023aaf5a32fa8423aec74ee9951200/review/reviewer/looking-for.md#style): ...

Quoting the abovementioned best practices:
> If you want to improve some style point that isn't in the style guide, prefix your comment with "Nit:" to let the developer know that it's a nitpick that you think would improve the code but isn't mandatory. Don't block CLs from being submitted based only on personal style preferences. [[read more](https://github.com/google/eng-practices/blob/bfe316ac6f023aaf5a32fa8423aec74ee9951200/review/reviewer/looking-for.md#style)]

I hope that's helpful.

thefuck/rules/git_branch_0v_to_dash_v.py
scorphus: > What are the advantages of making the code change on line 8?

The result of `'branch 0v'.split()` is exactly `['branch', '0v']`. In using the actual list, you're saving a few steps ([i.e.](https://www.merriam-webster.com/video/ie-vs-eg) calling a function that would split the string and return a list).

> Just looking at the code here on the PR and not in my editor, what I _think_ you're saying is that we can use a list comprehension to build a list of actions to take based on what the character after the erroneous `0`.

Yes. More precisely, that is a generator expression. It's better to use them in such cases as it stops at the first element found. The action would be the same for any of the characters after `0`, though: replace `0` with `-`.

> [...] Then I think you're saying that I can use the memoize annotation (is that what they're called in Python) to return the pre-defined "correct" and corrective action to take, thus allowing this functionality to be expanded upon bit by bit in future PRs.

You're partially correct regarding the memoization. There's a pattern for it in the project, as you can see in the link I posted. Let's take it as an example:

https://github.com/nvbn/thefuck/blob/365db1ee41d7bcd9f62d910f3b20361ea3499d92/thefuck/rules/git_add.py#L5-L27

Note that the `_get_missing_file` function is used in both `match` (on line 20) and `get_new_command` (on line 25). These are two separate *“phases”* of the *“correction”* process. TheFuck first tries to match a rule. If that rule matches, it then gets the new command. That means `_get_missing_file` is called two times. That's when the memoization comes in handy.

ps.: thinking over it again, maybe that's also a "Nit" if you consider the fact that the proposed function wouldn't be much expensive.

tests/rules/test_get_branch_0v_to_dash_v.py
scorphus: Yes! It's great that you followed patterns, that's a big yes. 👍 

But in that specific case, there's nothing special being done in the fixture. It simply returns an empty string. As you correctly suspected, it's more efficient (and also more readable) to just use the actual value (`""`) instead of the fixture.

tests/rules/test_get_branch_0v_to_dash_v.py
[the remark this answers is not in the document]
ProfessorTom: Yes, this was very helpful.

Is there an issue for linting? If not, can it be created/can I create it? (I suppose I _could_ go search; I'm really asking for permission here.)

tests/rules/test_get_branch_0v_to_dash_v.py
ProfessorTom: I tried replacing lines 6-8 with an empty string per your comment and that caused test failures. I tired just commenting out lines 6-8 and that lead to test failures too.

So, being unfamiliar with pytest fixtures, I assume these three lines need to stay in place.

If that is incorrect, please show me the correct way to remove this fixture and replace it with an empty string.

thefuck/rules/git_branch_0v_to_dash_v.py
ProfessorTom: I took the ideas presented here and made a run at implementation. Ultimately, I chose to use a dictionary with the flag as a key and a tuple of the "undo" plus the "what you meant to".

Re-requesting review; hopefully these changes will, at long last, allow this PR to get merged.

ProfessorTom:
@scorphus @nvbn Can I get a review on my latest changes?

ProfessorTom:
What will it take to get this new approach reviewed and merged? cc @nvbn @scorphus @jamtur01

scorphus:
Thank you for contributing, @ProfessorTom 👍 

Please check a new PR, which is probably going to be #1212.

ProfessorTom:
Thank you for merging.

PR #947, opened 2019-08-29
Change sudo.py pattern to lowercase

wx-csy:
The patterns in sudo.py should be in lowercase.

### DIFF
--- a/thefuck/rules/sudo.py
+++ b/thefuck/rules/sudo.py
@@ -21,7 +21,7 @@ patterns = ['permission denied',
             'edspermissionerror',
             'you don\'t have write permissions',
             'use `sudo`',
-            'SudoRequiredError',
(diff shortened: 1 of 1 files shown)


wx-csy:
I find that the "SudoRequiredError" only appears in [HostsCLI](https://github.com/dhilipsiva/hostscli). Since that tool is not used by many people I prefer completely removing all stuff related to that tool. It's better to provide these rules in a third-party package.

nvbn:
Looks good, thanks!

> I find that the "SudoRequiredError" only appears in HostsCLI. Since that tool is not used by many people I prefer completely removing all stuff related to that tool. It's better to provide these rules in a third-party package.

It's not making the rule noticeable slower and it might be useful for someone, so I guess it's fine to leave it.

PR #950, opened 2019-09-06
Correct "apt uninstall" -> "apt remove"

thatneat:
Previously, `apt uninstall` would correct to `apt install` but this is more of the desired behavior

I considered putting this in a separate rule, `apt_uninstall`, with a higher priority than `apt_invalid_operation` but this seemed like a more straightforward approach. If there's a preference for doing it that way I can do that instead.

### DIFF
--- a/tests/rules/test_apt_invalid_operation.py
+++ b/tests/rules/test_apt_invalid_operation.py
@@ -116,6 +116,8 @@ def test_get_operations(set_help, app, help_text, operations):
      apt_get_help, 'apt-get install vim'),
     ('apt saerch vim', invalid_operation('saerch'),
      apt_help, 'apt search vim'),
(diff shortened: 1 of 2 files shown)


ik1ne:
Running `flake8 (PROJECT_DIR)/thefuck/rules` gives following error:
```
thefuck/rules/apt_invalid_operation.py:4:1: F401 'thefuck.utils.replace_argument' imported but unused
```

I think removing that import will make PR pass the CI build.

thatneat:
oops, didn't see that this had failed. fixed!

nvbn:
Looks good, thanks!

nevercast:
Just the feature I was looking for, time to update my local `thefuck` copy! Thanks @thatneat

PR #961, opened 2019-10-02
Slow command timeout didn’t work

RetekBacsi:
+ Fixed debug message to include is_slow
* Using only the first word of shlex.split when checking if command is slow.


nvbn:
Thanks for the fix!

PR #963, opened 2019-10-03
Add fuck --hard as alternative to fuck --yeah

Faaab:
The one thing I always found counterintuitive to 'fuck --yeah' is that it is an option that I normally use when I get very frustrated. 'fuck --yeah' sounds much too positive for those occasions, which is why I think it would be great to be able to use 'fuck --hard'.

### DIFF
--- a/thefuck/argument_parser.py
+++ b/thefuck/argument_parser.py
@@ -55,7 +55,7 @@ class Parser(object):
         """It's too dangerous to use `-y` and `-r` together."""
         group = self._parser.add_mutually_exclusive_group()
         group.add_argument(
-            '-y', '--yes', '--yeah', '--hard'
+            '-y', '--yes', '--yeah', '--hard',
             action='store_true',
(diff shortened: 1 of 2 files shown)


thefuck/argument_parser.py
```diff
@@ -55,7 +55,7 @@ def _add_conflicting_arguments(self):
         """It's too dangerous to use `-y` and `-r` together."""
         group = self._parser.add_mutually_exclusive_group()
         group.add_argument(
-            '-y', '--yes', '--yeah',
+            '-y', '--yes', '--yeah', '--hard'
```
lomckee:
```suggestion
            '-y', '--yes', '--yeah', '--hard',
```

You missed a comma.

scorphus:
I also think it would be great to be able to use `fuck --hard`. Then, it would be nice to mention it in the README.

nvbn:
Thanks!

PR #966, opened 2019-10-07
added rule for drmemory commands missing --

elihschiff:
When running drmemory you need to add a `--` to the command. This rule helps add that.

so for example: `drmemory a.out` should be `drmemory -- a.out`

[http://www.drmemory.org/](http://www.drmemory.org/)

### DIFF
--- a/tests/rules/test_drmemory_no_app_specified.py
+++ b/tests/rules/test_drmemory_no_app_specified.py
@@ -1,0 +1,27 @@
+import pytest
+from thefuck.rules.drmemory_no_dashdash import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('command', [
+    Command('drmemory a.out', 'Usage: drmemory [options] --'),
+    Command('/etc/bin/drmemory test.exe', 'Usage: drmemory [options] --'),
+])
+def test_match(command):
+    assert match(command)
+
+
+@pytest.mark.parametrize('command', [
+    Command('drmemory -- a.out', ''),
+    Command('/etc/bin/drmemory -- test.exe', ''),
+])
+def test_not_match(command):
+    assert not match(command)
+
+
+@pytest.mark.parametrize('command, new_command', [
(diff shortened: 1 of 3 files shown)


tests/rules/test_drmemory_no_app_specified.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+from thefuck.rules.drmemory_no_dashdash import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('command', [
+    Command('drmemory a.out', 'Usage: drmemory [options] --'),
+    Command('/etc/bin/drmemory test.exe', 'Usage: drmemory [options] --'),
```
scorphus: Here, only `script` and `output` could be parametrized. Then in the function body `Command` can be then instantiated as in `assert match(Command(script, output))`.

Nit: Also, to reduce repetition a bit, a fixture `output` could be used.

Check [tests/rules/test_git_push.py](https://github.com/nvbn/thefuck/blob/master/tests/rules/test_git_push.py) for an example of both cases.

tests/rules/test_drmemory_no_app_specified.py
```diff
@@ -0,0 +1,27 @@
...
+    Command('drmemory -- a.out', ''),
+    Command('/etc/bin/drmemory -- test.exe', ''),
+])
+def test_not_match(command):
+    assert not match(command)
+
+
+@pytest.mark.parametrize('command, new_command', [
+    (Command('drmemory a.out', ''), 'drmemory -- a.out'),
+    (Command('/etc/bin/drmemory test.exe', ''), '/etc/bin/drmemory -- test.exe'),
```
scorphus: Dr. Memory's docs cites the following usage:

```
drmemory.exe -logdir c:/logs -- c:/path/to/my/app
```

The rule does not correctly fix the following wrong usage:

```
drmemory.exe -logdir c:/logs c:/path/to/my/app
```

PR #968, opened 2019-10-08
Support for yum invalid commands.

ik1ne:
This rule replaces incorrect yum commands, such as `yum isntall` into `yum install` and `yum remove` into `yum uninstall`.


nvbn:
Looks good, thanks!

Issue #975, opened 2019-10-11
Correcting `app-install` to `apt-get install` rather than `install`


yotam180:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.29 using Python 3.6.8 and Bash 4.4.20(1)-release

Your system (Debian 7, ArchLinux, Windows, etc.):

    Ubuntu 18.04.3 LTS

How to reproduce the bug:

    ~$ sudo apt-install python
    fuck

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

```
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': False,
 'no_colors': False,
 'num_close_matches': 3,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': PosixPath('/home/user/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
DEBUG: Received output: sudo: apt-install: command not found

[...]

PR #977, opened 2019-10-12
Add `wrong_hyphen_before_subcommand` rule

yotam180:
This rule corrects two common mistakes:

1. Omitting the space between the command and its flag arguments (e.g. `ls-la` instead of `ls -la`)
2. Placing a hyphen rather than a space between a command and its subcommand (e.g. `apt-install` and not `apt install`)

Regarding issue #975

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ following rules are enabled by default:
 * `vagrant_up` &ndash; starts up the vagrant instance;
 * `whois` &ndash; fixes `whois` command;
 * `workon_doesnt_exists` &ndash; fixes `virtualenvwrapper` env name os suggests to create new.
-* `wrong_hyphen_before_subcommand` &ndash; adds or omits dashes improperly placed (`ls-la` -> `ls -la`, `git-log` -> `git log`, etc.)
+* `wrong_hyphen_before_subcommand` &ndash; removes an improperly placed hyphen (`apt-install` -> `apt install`, `git-log` -> `git log`, etc.)
 * `yarn_alias` &ndash; fixes aliased `yarn` commands like `yarn ls`;
 * `yarn_command_not_found` &ndash; fixes misspelled `yarn` commands;
 * `yarn_command_replaced` &ndash; fixes replaced `yarn` commands;
--- a/README.md
+++ b/README.md
@@ -336,6 +336,7 @@ following rules are enabled by default:
 * `vagrant_up` &ndash; starts up the vagrant instance;
 * `whois` &ndash; fixes `whois` command;
 * `workon_doesnt_exists` &ndash; fixes `virtualenvwrapper` env name os suggests to create new.
+* `wrong_hyphen_before_subcommand` &ndash; adds or omits dashes improperly placed (`ls-la` -> `ls -la`, `git-log` -> `git log`, etc.)
 * `yarn_alias` &ndash; fixes aliased `yarn` commands like `yarn ls`;
 * `yarn_command_not_found` &ndash; fixes misspelled `yarn` commands;
 * `yarn_command_replaced` &ndash; fixes replaced `yarn` commands;
--- a/thefuck/rules/wrong_hyphen_before_subcommand.py
+++ b/thefuck/rules/wrong_hyphen_before_subcommand.py
@@ -90,7 +90,7 @@ def get_flags(man):
 
 def find_at_start(word, line):
     return line.startswith(word) and\
-        (len(line) == len(word) or line[len(word)] in ', ')
+        (len(line) == len(word) or line[len(word)] in ', (:')
 
 
 def find_subcommand(manpage, command, subcommand):
@@ -98,8 +98,8 @@ def find_subcommand(manpage, command, subcommand):
         any(map(partial(find_at_start, "%s-%s" % (command, subcommand)), manpage))
 
 
-def find_flag(manpage, flag):
-    return any(map(partial(find_at_start, flag), manpage))
+def find_flag(flags, flag):
+    return any(map(partial(find_at_start, '-' + flag), flags))
 
 
 @sudo_support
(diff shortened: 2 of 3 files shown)


yotam180:
I didn't abandon this PR to fail, currently working on mock-ups for the man-page retrieving functions.

thefuck/rules/wrong_hyphen_before_subcommand.py
```diff
@@ -0,0 +1,151 @@
...
+    manpage = get_manpage(cmd)
+    synopsis = get_synopsis(manpage)
+
+    if subcmd in synopsis or find_subcommand(manpage, cmd, subcmd):
+        # We are dealing with an accidentally added hyphen
+        return command.script.replace('-', ' ', 1)
+
+    else:
+        # We are dealing with a missing space
+        return command.script.replace('-', ' -', 1)
```
scorphus: Perhaps a simpler version of the entire rule would be better.

IMO, TheFuck should delegate that choice to the user.

Following the path the current rule is going, the failed command `apt-install foo` would get two suggestions, `apt -install foo` and `apt install foo`, among others.

Thing is that the former – `apt -install foo` – already gets suggested by the existing `missing_space_before_subcommand ` rule.

Looking at it that way, all the `wrong_hyphen_before_subcommand` – or even an *augmented* `missing_space_before_subcommand` – rule would have to do is replace the hyphen with a space.

What you think?

yotam180: Following your suggestions, I tried to simplify the rule down. I'm afraid I over-simplified it (since it may now suggest `git hello` as a replacement to `git-hello` even though `hello` is not a legitimate subcommand for `git`).

I'd be happy to hear your opinion about that. Should rules do above and beyond to verify their command suggestions? Or should the current way of command checking be the norm?

Anyway, another issue I have is that the rule now should not try to handle cases such as `ls-la` (-> `ls -la`), but as it doesn't check the manpages anymore there is no way for the rule to determine if `la` is a subcommand or flags. Thus it suggests `ls la`. Now the issue is that the priority of the rule (as I see it) should be lower than `no_command`'s priority, and the `missing_space_before_subcommand` is higher than `no_command`'s. Thus the `wrong_hyphen_before_subcommand` shadows the `missing_space_before_subcommand` rule.

Any suggestions?

scorphus: I think it's better if the rule does not try to be super smart. In the same sense as `missing_space_before_subcommand`. Let it suggest any fix that has some probability to work. The user decides to accept. That's my own opinion.

Regarding the `priority`, I think it should be set to `3900`. It's more close to `missing_space_before_subcommand` than to `no_command`.

@nvbn wanna chime in, pal?

README.md
```diff
@@ -299,6 +299,7 @@ following rules are enabled by default:
 * `vagrant_up` &ndash; starts up the vagrant instance;
 * `whois` &ndash; fixes `whois` command;
 * `workon_doesnt_exists` &ndash; fixes `virtualenvwrapper` env name os suggests to create new.
+* `wrong_hyphen_before_subcommand` &ndash; adds or omits dashes improperly placed (`ls-la` -> `ls -la`, `git-log` -> `git log`, etc.)
```
scorphus: The description lacks an update.

thefuck/rules/wrong_hyphen_before_subcommand.py
```diff
@@ -0,0 +1,22 @@
+from thefuck.utils import get_all_executables
+from thefuck.specific.sudo import sudo_support
+
+
+@sudo_support
+def match(command):
+    if '-' not in command.script_parts[0] or command.script_parts[0] in get_all_executables():
+        return False
+
+    cmd, subcmd = command.script_parts[0].split('-')[:2]
```
scorphus: Now that `subcmd` isn't necessary anymore, this line could be changed to:
```suggestion
    cmd = command.script_parts[0].split('-', 1)[0]
```

thefuck/rules/wrong_hyphen_before_subcommand.py
```diff
@@ -0,0 +1,22 @@
...
+    if cmd not in get_all_executables():
+        return False
+
+    return True
+
+
+@sudo_support
+def get_new_command(command):
+    return command.script.replace("-", " ", 1)
```
davidhart82: This is dangerous, there might be other hyphens. You really should only replace the hyphen in script_parts[0].

thefuck/rules/wrong_hyphen_before_subcommand.py
```diff
@@ -0,0 +1,22 @@
...
+from thefuck.specific.sudo import sudo_support
+
+
+@sudo_support
+def match(command):
+    if '-' not in command.script_parts[0] or command.script_parts[0] in get_all_executables():
+        return False
+
+    cmd, subcmd = command.script_parts[0].split('-')[:2]
+    if cmd not in get_all_executables():
```
davidhart82: This could simply be :

```python
return cmd in get_all_executables()
```

thefuck/rules/wrong_hyphen_before_subcommand.py
scorphus: I would agree if the `match` function wouldn't require a hyphen in `command.script_parts[0]`.

thefuck/rules/wrong_hyphen_before_subcommand.py
scorphus: @yotam180, can you please address this remark?
```suggestion
    return cmd in get_all_executables()
```

This along with [the one a few boxes above](https://github.com/nvbn/thefuck/pull/977#discussion_r336793322):
```diff
-    cmd, subcmd = command.script_parts[0].split('-')[:2]
+    cmd = command.script_parts[0].split('-', 1)[0]
```

scorphus (approved):
Thanks for ccontributing, @yotam180! 👍 :shipit: 

You may want to check the last few commits.

gjule040-code (approved):

Issue #965, opened 2019-10-06
Instructions in contributing.md don't work, should be replaced with pip3 install . develop


jaredgoodman03:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.29 using Python 3.6.8 and Bash 4.4.20(1)-release

Your system (Debian 7, ArchLinux, Windows, etc.):

    Linux Mint 19.2 Tina

How to reproduce the bug:
Running `python3 setup.py` develop as explained in `CONTRIBUTING.md` doesn't actually install the development version on the system. I found better luck by running `sudo pip3 install . develop` in the root directory of the project. Also, [this](https://stackoverflow.com/questions/28210269/importerror-cannot-import-name-main-when-running-pip-version-command-in-windo) needed to be changed.

Anything else you think is relevant:

    Should I change CONTRIBUTING.md to reflect this and submit a PR, or is this just something wacky on my machine?

scorphus:
That's great you wish to contribute! Did you try with a virtual environment activated?

As with almost any Python development project, a virtual environment is expected to be created and activated first. And that very much depends on how you as a Python developer create and manage virtual environments ([pyenv](https://github.com/pyenv/pyenv#installation) in my case). Something like that could then be a good addition to [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md). What you think?

steaksauce-:
On some distributions, Python 2.7 is still the default, and thus `pip` will default to Python 2 version. As @scorphus mentions, you should activate your venv first. If your venv is based on python 3, `pip` will be the python 3 version istead of 2.

scorphus:
Fix on its way in #976

deongdem:
hello there! are you having problems with installing module's then here's your answer ---
i've used linux system i also such problems so i tried pip3 install <module name> if it is not working then contact me deongdem4@gmail.com ----- my email address

scorphus:
Fixed in 70b414aca2acbbac2cd3c75bc4c3e2dd489b3144.

PR #976, opened 2019-10-11
Issue#965 - added venv instructions to CONTRIBUTING.md

shawnmcgraw:
Added instructions in CONTRIBUTING.md for creating python3 venv. Hopefully this will reduce alleviate the confusion that lead to the creation of this issue. Let me know if you would like anything changed. I'll be glad to help.

PR #978, opened 2019-10-14
Add instructions for virtualenv in CONTRIBUTING.md

kain88-de:
Fix:  #965


scorphus:
Thanks for your intent to contribute. But this is a duplicate of #976. If you think it should be tackled differently, please add a comment to #976.

PR #880, opened 2019-01-20
Added korean keyboard layout

yangkyeongmo:
Not only layout of korean keyboard but decomposing korean letters is also added.
If korean letter is written, it's automatically combined into one letter. ex) ㄱ + ㅏ = 가
So decomposing it was necessary.

PR #981, opened 2019-10-16
Add support for switch_lang for Korean.

ik1ne:
Updated PR of #880 .


thefuck/rules/switch_lang.py
```diff
@@ -63,15 +77,43 @@ def _switch_command(command, layout):
...
+                    hg_str += _change_double(ch)
+        else:
+            hg_str += _change_double(ch)
+    return hg_str
+
+
 def match(command):
     if 'not found' not in command.output:
         return False
+    if any([u'ㄱ' <= ch <= u'ㅎ' or u'ㅏ' <= ch <= u'ㅣ' or u'가' <= ch <= u'힣'
```
scorphus: If you remove the square brackets here, you'll be passing a generator to `any` instead of a list. That way, the iteration stops at the very first element of `command.script` that meets the conditional expression, not having to iterate over all of them for every case. Generators FTW! ✨

thefuck/rules/switch_lang.py
```diff
@@ -63,15 +77,43 @@ def _switch_command(command, layout):
...
+            for ch in command.script]):
+        return True
 
     matched_layout = _get_matched_layout(command)
     return (matched_layout and
             _switch_command(command, matched_layout) != get_alias())
 
 
 def get_new_command(command):
+    if any([u'ㄱ' <= ch <= u'ㅎ' or u'ㅏ' <= ch <= u'ㅣ' or u'가' <= ch <= u'힣'
```
scorphus: Ditto.

thefuck/rules/switch_lang.py
```diff
@@ -63,15 +77,43 @@ def _switch_command(command, layout):
     return ''.join(_switch(ch, layout) for ch in command.script)
 
 
+def _decompose_korean(command):
+    def _change_double(ch):
+        if ch in DOUBLE_LIST:
+            return DOUBLE_MOD_LIST[DOUBLE_LIST.index(ch)]
+        else:
+            return ch
```
scorphus: Nit: this `else` statement is unnecessary, as its `if` counterpart has a `return` in its block (on line 83). So it can be removed and the `return ch` statement can be stepped one indentation back. That's for improved readability.

```suggestion
        return ch
```

ik1ne:
Done, I also changed `_switch` function which was not written by me, but for consistency's sake I did it anyways.

scorphus:
@nvbn This contribution has two authors. I'm wondering if we keep authorship in the commits or not. How should we proceed? Combine commits into one per author? Or merge them all into one?

nvbn:
Looks good, thanks!

nvbn:
@scorphus I thought that github can handle that, added everyone in the commit message just in case =)

PR #986, opened 2019-10-21
Add pyenv to gitignore

aaditkamat:
I have started using `pyenv` to create and manage Python virtual environments. I noticed that there is no config for `pyenv` within the current gitignore so I thought I would add it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,5 +59,8 @@ target/
 .env
 .idea
 
+# pyenv
+.python-version/
+
 # vim temporary files
 .*.swp


scorphus:
Everything that is not specific and directly related to the project should be kept in a separate, global ignore file, e.g. `$HOME/.config/git/ignore`. See `man gitignore | grep -A5 'all situations'`:

       o   Patterns which a user wants Git to ignore in all situations (e.g.,
           backup or temporary files generated by the user's editor of choice)
           generally go into a file specified by core.excludesFile in the
           user's ~/.gitconfig. Its default value is
           $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set
           or empty, $HOME/.config/git/ignore is used instead.

aaditkamat:
This is a project specific configuration though

scorphus:
No. This is actually related to Python development in general, not to this specific project. If most of your Python projects have a `.python-version` entry in `.gitignore`, that entry should be global instead. Also, other developers might prefer to use another entirely different tool to manage their virtual environments, in which case that entry won't make any sense.

PR #988, opened 2019-10-24
Set rm / to be a safe call

### LINKED ISSUE #943: Detect derivates of the rm  /
Fuck should not 'help' us with removing our /, but warn us in about it

yoavtzelnick:
Closes #943

### DIFF
--- a/tests/rules/test_rm_root.py
+++ b/tests/rules/test_rm_root.py
@@ -4,17 +4,15 @@ from thefuck.types import Command
 
 
 def test_match():
-    assert match(Command('rm -rf /', 'add --no-preserve-root'))
+    assert match(Command('rm -rf /', ''))
 
 
 @pytest.mark.parametrize('command', [
     Command('ls', 'add --no-preserve-root'),
-    Command('rm --no-preserve-root /', 'add --no-preserve-root'),
-    Command('rm -rf /', '')])
+    Command('rm', '/usr/bin/python')])
 def test_not_match(command):
     assert not match(command)
 
 
 def test_get_new_command():
-    assert (get_new_command(Command('rm -rf /', ''))
-            == 'rm -rf / --no-preserve-root')
+    assert 'rm' not in (get_new_command(Command('rm -rf /', '')))
--- a/thefuck/rules/rm_root.py
+++ b/thefuck/rules/rm_root.py
@@ -6,11 +6,9 @@ enabled_by_default = False
 @sudo_support
 def match(command):
     return (command.script_parts
-            and {'rm', '/'}.issubset(command.script_parts)
-            and '--no-preserve-root' not in command.script
-            and '--no-preserve-root' in command.output)
+            and {'rm', '/'}.issubset(command.script_parts))
 
 
 @sudo_support
 def get_new_command(command):
-    return u'{} --no-preserve-root'.format(command.script)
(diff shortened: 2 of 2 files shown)


scorphus:
This change makes no sense, IMHO. That is the same as not having this rule or disabling it – which is, a matter of fact, the default. The rule is disabled by default.

The user that has it enabled, did so for a reason.

yoavtzelnick:
I disagree, actually notifying the person automatically writing `fuck`, to notice what they're writing as @steaksauce suggested is important IMO

Dzordzu:
Personally, I absolutely support the idea of informing people about potential danger. But disabling (without special option that will allow to do such a thing) is slightly far-fetched.

scorphus:
Agreed.

So, perhaps a `side_effect` is a fair option. Something like:

```python
import sys
try:
    input = raw_input  # Python 2
except NameError:
    pass
# ...
def side_effect(old_cmd, command):
    warn("DANGER!!! THIS MAY DESTROY YOUR SYSTEM! ARE YOU SURE? (y/N): ")
    reply = input().strip().lower()  # raw_input should be used in Python 2
    if not reply or reply[0] != "y":
        sys.exit(0)
```

What do you all think?

Dzordzu:
This satisfies me for this moment . Thank you :)

scorphus:
What do you say, @yoavtzelnick? 🙂

tests/rules/test_rm_root.py
```diff
@@ -1,20 +1,14 @@
...
-    Command('rm --no-preserve-root /', 'add --no-preserve-root'),
-    Command('rm -rf /', '')])
+    Command('rm', '/usr/bin/python')])
 def test_not_match(command):
     assert not match(command)
-
-
-def test_get_new_command():
-    assert (get_new_command(Command('rm -rf /', ''))
-            == 'rm -rf / --no-preserve-root')
```
scorphus: Existing tests shouldn't be touched, as there's only one new function added to the rule, namely `side_effect`. Please include tests for it.

thefuck/rules/rm_root.py
```diff
@@ -6,11 +8,19 @@
...
-            and {'rm', '/'}.issubset(command.script_parts)
-            and '--no-preserve-root' not in command.script
-            and '--no-preserve-root' in command.output)
+            and {'rm', '/'}.issubset(command.script_parts))
 
 
-@sudo_support
 def get_new_command(command):
-    return u'{} --no-preserve-root'.format(command.script)
+    return '{} {} {}'.format(command.script_parts[0], '-rf', ' '.join(command.script_parts[1:]))
```
scorphus: The way it is, the rule is not doing what it is supposed to do. All these changes should be removed from the CL. Only the `side_effect` function should be part of it.

thefuck/rules/rm_root.py
```diff
@@ -6,11 +8,19 @@
...
+
+
+def side_effect(old_cmd, command):
+    warn("DANGER!!! THIS MAY DESTROY YOUR SYSTEM! ARE YOU SURE? (y/N): ")
+    try:
+        input = raw_input
+    except NameError:
+        pass
+    reply = input().strip().lower()  # raw_input should be used in Python 2
```
scorphus: Nit: that comment could be moved to line 21.

PR #991, opened 2019-10-25
Added .vscode to gitignore for people editing with VSCode.

BubbsTheSupreme:
I use VSCode to program as I'm sure others do and noticed that the VSCode file wasn't ignored so I thought it would be helpful to add it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ target/
 
 # vim temporary files
 .*.swp
+
+# Editor specific files
+.vscode/


.gitignore
```diff
@@ -61,3 +61,6 @@ target/
 
 # vim temporary files
 .*.swp
+
+# Editor specific files
+.vscode/
```
SuperSandro2000: missing new line

scorphus:
Visual Studio Code is *your* editor of choice. There are several other editors out there. Everything that is not specific and directly related to the project should be kept in a separate, global ignore file, e.g. `$HOME/.config/git/ignore`. See `man gitignore | grep -A5 'all situations'`:

       o   Patterns which a user wants Git to ignore in all situations (e.g.,
           backup or temporary files generated by the user's editor of choice)
           generally go into a file specified by core.excludesFile in the
           user's ~/.gitconfig. Its default value is
           $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set
           or empty, $HOME/.config/git/ignore is used instead.

BubbsTheSupreme:
I understand, wont happen again

Issue #956, opened 2019-09-22
Suggest correct filename for editing


marv3lls:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

   The Fuck 3.29 using Python 3.6.8 and ZSH 5.4.2

Your system (Debian 7, ArchLinux, Windows, etc.):

    Ubuntu 18.04.3 LTS

How to reproduce the bug:

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

```

Script started on 2019-09-21 18:21:17-0700
]1337;RemoteHost=root@marvell.cat]1337;CurrentDir=/home/she/http/swo.re/includes]1337;ShellIntegrationVersion=6;shell=zsh[1m[7m#[27m[1m[0m                                                             
 
]2;root@marvell: /home/she/http/swo.re/includes]1;..o.re/includes]133;D;0]1337;RemoteHost=root@marvell.cat]1337;CurrentDir=/home/she/http/swo.re/includes
[0m[27m[24m[J]133;A[32m18:21:18[00m [34mroot[00m@marvell.cat[00m [34mincludes[00m ]133;B[K[?1h=[?2004h[32mscript[39m [4m~/script.txt[24m[19D[32mf[32mu[32mc[32mk[39m[39m [39m  [24m [24m [24m [24m [24m [24m [24m [24m [24m [24m [24m [24m [15D[?1l>[?2004l

]2;fuck]1;fuck]133;C;Seems like [1mfuck[0m alias isn't configured!
Please put [1meval $(thefuck --alias)[0m in your [1m~/.zshrc[0m and apply changes with [1msource ~/.zshrc[0m or restart your shell.
Or run [1mfuck[0m second time for configuring it automatically.
More details - https://github.com/nvbn/thefuck#manual-installation
[0m[1m[7m#[27m[1m[0m                                                             
 
]2;root@marvell: /home/she/http/swo.re/includes]1;..o.re/includes]133;D;0]1337;RemoteHost=root@marvell.cat]1337;CurrentDir=/home/she/http/swo.re/includes
[...]

PR #992, opened 2019-10-25
Add edit filename rule

riley-martine:
Closes https://github.com/nvbn/thefuck/issues/956

Will need code review - I'm not familiar with your codebase.

Questions:
- Any glaring test omissions?
- Am I doing the right way to return multiple from get_new_command?
- I don't like the rule file I made - am I misunderstanding match?

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ following rules are enabled by default:
 * `docker_not_command` &ndash; fixes wrong docker commands like `docker tags`;
 * `docker_image_being_used_by_container` &dash; removes the container that is using the image before removing the image;
 * `dry` &ndash; fixes repetitions like `git git push`;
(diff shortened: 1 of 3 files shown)


davidhart82:
I might have done something like this:

```python
import os
from thefuck import utils


def match(command):
    return command.script_parts[0] in ['nano', 'vi', 'less'] or 'No such file or directory' in command.output


def get_new_command(command):
    path = os.path.dirname(os.path.join(os.getcwd(), command.script_parts[-1]))

    closest_file = utils.get_closest(command.script_parts[-1], os.listdir(path))
    if closest_file:
        return utils.replace_argument.(command.script, command.script_parts[-1], closest_file)


requires_output = False
priority = 3100
```

davidhart82:
Maybe even use `get_close_matches` instead of `get_closest`.

Issue #946, opened 2019-08-25
Only including necessary envs in THEFUCK_DEBUG=true


SuperSandro2000:
Thefuck shouldn't include all env variables in the debug output. It is not necessary that you have private tokens, ssh connection ips and other personal identifiable information.

The following should be enough
````
TF_SHELL=bash
TF_ALIAS=c
TF_SHELL_ALIASES=alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'
````

The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.29 using Python 3.7.3 and Bash 5.0.3(1)-release

Your system (Debian 7, ArchLinux, Windows, etc.):

    Ubuntu 19.04

How to reproduce the bug:

    Run thefuck with THEFUCK_DEBUG=true enabled.

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

    Not applicable

haneybarg:
Hi! Can I take this?

SuperSandro2000:
@haneybarg Just do it...

PR #995, opened 2019-10-26
Only including necessary envs in THEFUCK_DEBUG=true.

haneybarg:
Issue: #946.

Reason:
`Thefuck shouldn't include all env variables in the debug output. It is not necessary that you have private tokens, ssh connection ips and other personal identifiable information.`

Modifications:
`Kept the following three env variables only:
- TF_SHELL=bash
- TF_ALIAS=c
- TF_SHELL_ALIASES`

### DIFF
--- a/thefuck/output_readers/rerun.py
+++ b/thefuck/output_readers/rerun.py
@@ -52,11 +52,12 @@ def get_output(script, expanded):
     """
     env = dict(os.environ)
     env.update(settings.env)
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
> Even then, I'd need to give it more thought.

It would be a big improvement. I use fzf and this also adds a lot of envs and my default debug output is very long and filled with stuff I don't want to share with the world.

PR #996, opened 2019-10-28
Add rule to remove leading shell prompt literal $

boonwj:
Rule added to remove leading shell prompt literals found in the command.

This usually happens when a command is copy pasted from documentations with code blocks that contain it.

```
$ $ echo hello world
bash: $: command not found...
$ fuck
echo hello world [enter/↑/↓/ctrl+c]
hello world
```

### DIFF
--- a/tests/rules/test_remove_shell_prompt_literal.py
+++ b/tests/rules/test_remove_shell_prompt_literal.py
@@ -3,30 +3,31 @@ from thefuck.rules.remove_shell_prompt_literal import match, get_new_command
 from thefuck.types import Command
 
 
-@pytest.mark.parametrize('command',
+@pytest.mark.parametrize(
+    "command",
     [
-        Command('$ cd newdir', '$: command not found'),
-        Command(' $ cd newdir', '$: command not found'),
-    ])
+        Command("$ cd newdir", "$: command not found"),
+        Command(" $ cd newdir", "$: command not found"),
+    ],
+)
 def test_match(command):
     assert match(command)
 
 
-@pytest.mark.parametrize('command',
-    [
-        Command('$', ''),
-        Command('$?', ''),
-        Command(' $?', ''),
-        Command('', ''),
-    ])
+@pytest.mark.parametrize(
+    "command",
+    [Command("$", ""), Command("$?", ""), Command(" $?", ""), Command("", "")],
+)
 def test_not_match(command):
     assert not match(command)
 
 
-@pytest.mark.parametrize('command, new_command',
+@pytest.mark.parametrize(
+    "command, new_command",
     [
-        (Command('$ cd newdir', ''), 'cd newdir'),
-        (Command('$ python3 -m virtualenv env', ''), 'python3 -m virtualenv env'),
-    ])
+        (Command("$ cd newdir", ""), "cd newdir"),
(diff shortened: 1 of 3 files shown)


tests/rules/test_remove_shell_prompt_literal.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+from thefuck.rules.remove_shell_prompt_literal import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize(
+    "command",
```
scorphus: How about parametrizing only `script` (as string) and declare a `Command` instance with `"$: command not found"` as `output`?

```suggestion
    "script",
```

That would make all test tables more readable.

tests/rules/test_remove_shell_prompt_literal.py
```diff
@@ -0,0 +1,33 @@
...
+    [
+        Command("$ cd newdir", "$: command not found"),
```
scorphus:
```suggestion
        "$ cd newdir",
```

tests/rules/test_remove_shell_prompt_literal.py
```diff
@@ -0,0 +1,33 @@
...
+        Command(" $ cd newdir", "$: command not found"),
+    ],
+)
+def test_match(command):
```
scorphus:
```suggestion
def test_match(script):
```

tests/rules/test_remove_shell_prompt_literal.py
```diff
@@ -0,0 +1,33 @@
...
+    assert match(command)
```
scorphus:
```suggestion
    assert match(Command(script, "$: command not found"))
```

tests/rules/test_remove_shell_prompt_literal.py
```diff
@@ -0,0 +1,33 @@
...
+
+@pytest.mark.parametrize(
+    "command, new_command",
+    [
+        (Command("$ cd newdir", ""), "cd newdir"),
+        (Command("$ python3 -m virtualenv env", ""), "python3 -m virtualenv env"),
+    ],
+)
+def test_get_new_command(command, new_command):
+    assert get_new_command(command) == new_command
```
scorphus: Ditto.

thefuck/rules/remove_shell_prompt_literal.py
```diff
@@ -0,0 +1,25 @@
...
+        "$: command not found" in command.output
+        and re.search(r"^[\s]*\$ [\S]+", command.script) is not None
+    )
+
+
+def get_new_command(command):
+    return command.script.replace("$", "", 1).strip()
+
+
+requires_output = True
```
scorphus: This can be omitted as [`requires_output`'s default value is `True`](https://github.com/nvbn/thefuck/blob/70b414aca2acbbac2cd3c75bc4c3e2dd489b3144/thefuck/types.py#L147).

boonwj:
Thanks for the comments @scorphus! 

I have made the modifications as suggested. However, the non matches were not changed as I realised that the outputs were not the same for every test case.

nvbn:
Looks good, thanks!

PR #998, opened 2019-10-29
Add choco_install rule

philiparola:
Adds a rule to append '.install' to a chocolatey install command that
failed because of a non-existent package.

### DIFF
--- a/tests/rules/test_choco_install.py
+++ b/tests/rules/test_choco_install.py
@@ -1,0 +1,12 @@
+import pytest
+from thefuck.rules.choco_install import get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('before, after', [
+    ('choco install logstitcher', 'choco install logstitcher.install'),
+    ('cinst logstitcher', 'cinst logstitcher.install'),
+    ('choco install logstitcher -y', 'choco install logstitcher.install -y'),
+    ('cinst logstitcher -y', 'cinst logstitcher.install -y')])
+def test_get_new_command(before, after):
+    assert (get_new_command(Command(before, '')) == after)
--- a/tests/rules/test_choco_install.py
+++ b/tests/rules/test_choco_install.py
@@ -1,6 +1,74 @@
 import pytest
-from thefuck.rules.choco_install import get_new_command
+from thefuck.rules.choco_install import match, get_new_command
 from thefuck.types import Command
+
+
+package_not_found_error = (
+    'Chocolatey v0.10.15\n'
+    'Installing the following packages:\n'
+    'logstitcher\n'
+    'By installing you accept licenses for the packages.\n'
+    'logstitcher not installed. The package was not found with the source(s) listed.\n'
+    ' Source(s): \'https://chocolatey.org/api/v2/\'\n'
+    ' NOTE: When you specify explicit sources, it overrides default sources.\n'
+    'If the package version is a prerelease and you didn\'t specify `--pre`,\n'
+    ' the package may not be found.\n'
+    'Please see https://chocolatey.org/docs/troubleshooting for more\n'
+    ' assistance.\n'
+    '\n'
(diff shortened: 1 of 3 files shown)


tests/rules/test_choco_install.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from thefuck.rules.choco_install import get_new_command
+from thefuck.types import Command
+
+
```
scorphus: Tests for `match` are missing. Check other tests for reference.

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
+def match(command):
+    return (('choco install' in command.script_parts or 'cinst' in command.script_parts)
+            and 'Installing the following packages' in command.output)
+
+
+def get_new_command(command):
+    cmdList = command.script.split(' ')
```
scorphus: `command.script_parts` should be used instead.
```suggestion
```
See below.

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
...
+    packageName = ""
+    # Find the argument that is the package name
+    for i in cmdList:
```
scorphus: `command.script_parts` should be used instead.

```suggestion
    for script_part in command.script_parts:
```

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
+def match(command):
+    return (('choco install' in command.script_parts or 'cinst' in command.script_parts)
+            and 'Installing the following packages' in command.output)
+
+
+def get_new_command(command):
+    cmdList = command.script.split(' ')
+    packageName = ""
```
scorphus:
```suggestion
```

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
...
+    # Find the argument that is the package name
+    for i in cmdList:
+        print(i)
```
scorphus:
```suggestion
```

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
...
+        if "choco" in i:
```
scorphus: This `if` statement sequence should be rewritten with all expressions grouped together with `and` operators.

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
+def match(command):
```
scorphus:
```suggestion
from thefuck.utils import for_app


@for_app("choco", "cinst")
def match(command):
```

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
...
+    return (('choco install' in command.script_parts or 'cinst' in command.script_parts)
```
scorphus:
```suggestion
    return ((command.script.startswith('choco install') or 'cinst' in command.script_parts)
```

Also, the outer parenthesis can be removed.

thefuck/rules/choco_install.py
```diff
@@ -0,0 +1,29 @@
...
+        if '=' in i:            # Some paramaters contain '='
+            continue
+        if '/' in i:            # Some parameters contain slashes
+            continue
+        else:
+            packageName = i
+    # Find the name of the broken package, and append metapackage names
+    if not packageName:
+        return False
+    return(command.script.replace(packageName, packageName + ".install"))
```
scorphus: Should be disabled when those commands are not available.
```suggestion
    return(command.script.replace(packageName, packageName + ".install"))


enabled_by_default = bool(which("choco")) or bool(which("cinst"))
```

scorphus:
Thanks for your contribution, @philiparola!

nvbn:
Looks good, thanks!

PR #1000, opened 2019-10-31
Update requirements.txt

drishyasankar:


scorphus:
What's the reason behind it. What breaks with the 17.1 version? I see no option other than closing this.

PR #1002, opened 2019-10-31
Update release.py

drishyasankar:


scorphus:
This CL does not improve the code base, release.py is not even part of it. IMO it can be closed.

VilhelmPrytz:
@scorphus Agreed that it doesn't improve the code base, but it still does some sort of improvement. Merging is just a click away :smile: 

Cheers

PR #1004, opened 2019-10-31
fix: incorrect powershell alias instruction

yume-chan:
Just setup PowerShell on my new MacBook and when I run `fuck` it says:

image

I tried, but `iex "thefuck --alias"` will only print out the alias script.

The [wiki page](https://github.com/nvbn/thefuck/wiki/Shell-aliases#powershell) is correct.

I'm so confused that the typo is there from the very beginning, and for about 3 years no one had found it.

EDIT: And use `& $PROFILE` to reload profile is also incorrect because the `&` (call operator) execute the script in a child scope, so any function definition will not expose to the outer (main running) scope. We should use the `.` (dot-sourcing) operator instead. See this [StackOverflow answer](https://stackoverflow.com/a/54680512).

### DIFF
--- a/thefuck/shells/powershell.py
+++ b/thefuck/shells/powershell.py
@@ -26,7 +26,7 @@ class Powershell(Generic):
         return ShellConfiguration(
             content=u'iex "$(thefuck --alias)"',
             path='$profile',
-            reload='& $profile',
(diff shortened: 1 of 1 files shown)


nvbn:
Looks good, thanks!

PR #1006, opened 2019-10-31
Fallback to su if sudo doesn't exist

moore3071:
su should exist even if sudo doesn't. As such, we can try to run a command with su if sudo doesn't exist.

### DIFF
--- a/thefuck/rules/su.py
+++ b/thefuck/rules/su.py
@@ -1,0 +1,15 @@
+def match(command):
+    if command.script_parts and '&&' not in command.script_parts and command.script_parts[0] == 'su':
+        return False
+
+    return 'command not found: sudo' in command.output.lower()
+
+
+def get_new_command(command):
+    if '&&' in command.script:
+        return u'su -c "sh -c "{}""'.format(" ".join([part for part in command.script_parts if part != "sudo"]))
+    elif '>' in command.script:
+        return u'su -c "sh -c "{}""'.format(command.script.replace('"', '\\"'))
+    else:
+        return u'su -c {}'.format(command.script)
+priority = 1200


scorphus:
Thanks for contributing. Can you please add tests?

moore3071:
Will do. It might not be until next week though.

tests/rules/test_su.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+from thefuck.rules.su import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('output', [
+    'command not found: sudo'])
+def test_match(output):
+    assert match(Command('', output))
```
scorphus: It would be good having some practical test cases. `command.script_parts` is part of the `match` condition but not taken into consideration in tests.

tests/rules/test_su.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_not_match():
+    assert not match(Command('', ''))
+    assert not match(Command('sudo ls', 'Permission denied'))
+    assert not match(Command('su -c ls', 'Permission denied'))
+    assert not match(Command('ls', 'command not found: ls'))
```
scorphus: `@ pytest.mark.parametrize` is being used in `test_match` for one entry and not in `test_not_match` for four entries. Use it here as well.

thefuck/rules/su.py
```diff
@@ -0,0 +1,12 @@
+def match(command):
```
scorphus: `thefuck.utils.for_app` is there for the help \o/

```suggestion
from thefuck.utils import for_app


@for_app('sudo')
def match(command):
```

thefuck/rules/su.py
```diff
@@ -0,0 +1,12 @@
...
+    if command.script_parts and '&&' not in command.script_parts and command.script_parts[0] == 'su':
```
scorphus:
```suggestion
    if '&&' not in command.script_parts:
```

Also, Fish Shell must be considered. Support to `&&` operator has been added to version 3. Earlier versions only support the `and` operator.

djh82: sudo.py suffers with the same issue, and that's already released.

scorphus: That's true. I's been sitting there unnoticed. Thanks, @djh82. That must be fixed.

PR #1013, opened 2019-11-27
fix missing_space_before_subcommand.py rule for apt-get

delamemily1:
Prior to my changes, the missing_space_before_subcommands rule looped through the available commands, checked to see if the first portion of the command started with a valid executable, and then returned that executable immediately if it did. Because of this, it always converted **apt-get** commands into **apt     -get** commands, and other rules of the like. I changed it under recommendation of @scorphus that it return all possible matching missing space commands, not just the first one.

Here is a screen shot of this **buggy behavior prior to my changes:**

command: execsnoopcats [incorrect behavior- only returns the first instance: ex ecsnoopcats]
Screen Shot 2019-12-02 at 10 09 19 AM

Here is a screenshot of the **correct behavior after my changes:**

command: execsnoopcats [correct behavior - returning both ex ecsnoopcats and execsnoop cats]
Screen Shot 2019-12-02 at 10 08 04 AM
Screen Shot 2019-12-02 at 10 07 56 AM

### DIFF
--- a/thefuck/rules/missing_space_before_subcommand.py
+++ b/thefuck/rules/missing_space_before_subcommand.py
@@ -7,11 +7,12 @@ def _get_executable(script_part):
         if script_part.startswith(executable):
             if executable.startswith("apt"):
                 if script_part.startswith("apt-get"):
-                    if len(script_part)> len("apt-get"):
+                    if len(script_part) > len("apt-get"):
                         return "apt-get"
                     else:
                         return
             return executable
+
 
 def match(command):
     return (not command.script_parts[0] in get_all_executables()
@@ -24,11 +25,3 @@ def get_new_command(command):
 
 
 priority = 4000
-
-
-
-
-
-
-
-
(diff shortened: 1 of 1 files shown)


delamemily1:
**my continuous-integration tests only fail for python2.7** and they don't appear to be erroring out on my rule but on a different rule. also, since python2.7 will no longer support thefuck, **is it possible that my branch is still merged despite the continuous integration tests failing with python2.7?**

scorphus:
As for the build error, it's an issue introduced by latter releases of [pytest-mock](/pytest-dev/pytest-mock) and is already being taken care of.

delamemily1:
> This feature is already covered by the `missing_space_before_subcommand` rule.

I know, this is a correction to the 'missing_space_before_subcommand' rule because it always early exits on apt, making the check for a missing space on an apt-get command never take place. my change ensures that it differentiates between a missing space on an "apt" command and an "apt-get" command

delamemily1:
will do i did not realize how many apt- commands there were lol

PR #1020, opened 2019-12-09
Feature/self protect thefuck

delamemily1:


davidhart82:
This seems frivolous? And only actually applies if you try to uninstall incorrectly. There are other rules which actually correct the command.

PR #1022, opened 2019-12-11
Allow multiple returns from git_checkout

davidhart82:


davidhart82:
It doesn't look like my changes have caused the build failure?

thefuck/rules/git_checkout.py
```diff
@@ -34,10 +34,17 @@ def get_new_command(command):
...
-    elif command.script_parts[1] == 'checkout':
-        return replace_argument(command.script, 'checkout', 'checkout -b')
+        new_commands.append(replace_argument(command.script, missing_file, closest_branch))
+    if command.script_parts[1] == 'checkout':
+        new_commands.append(replace_argument(command.script, 'checkout', 'checkout -b'))
+
+    if new_commands:
+        return new_commands
     else:
+        # if neither the better options above match, then resort to this
```
davidhart82: Now I've implemented this, I wonder whether to simply add this to the new_commands list and be done with it?

davidhart82: Maybe not, I think I've misunderstood what this last correction was doing!

davidhart82:
I think that the match is also wrong for git_checkout; my version of git (2.24.0) doesnt have the full stop at the end.

nvbn:
Looks good, thanks!

The ci jobs failures weren't related to the changes.

PR #1023, opened 2019-12-14
fix: removed useless redefinition of "path" variable

elihschiff:


nvbn:
Thanks for noticing and fixing!

PR #1024, opened 2019-12-15
Added missing dependencies for py.test to req.txt

x-zvf:
Without there dependencies, which are not installed with the original
requirements.txt, on a fresh python 3.8 venv (on GNU/Linux), py.test
will not work.


nvbn:
You need to also run `python setup.py develop` as in https://github.com/nvbn/thefuck/blob/master/CONTRIBUTING.md#developing and it will work.

Issue #881, opened 2019-01-24
Doesn't repair apt-get command


thiswillbeyourgithub:
The Fuck 3.28 using Python 3.6.7 and ZSH 5.4.2
Ubuntu 18.04 LTS

**How to reproduce the bug:**

type `apt-get insatll test` which is a wrong command ("insatll" insteaf of install)
returns `E: Invalid operation insatl`
type `fuck`
returns a previous apt-get command : `sudo apt-get install redshift [enter/↑/↓/ctrl+c]`
what should happen is a correction of insatll into install, as well as for at-get apt-gt etc

davidhart82:
This occurs because the help content of 'apt-get' now looks like the help content for 'apt' and so when the apt_invalid_operation rule tries to parse it to find the appropriate replacement, it fails.

PR #1031, opened 2020-01-05
Apt-get help is now much more like apt

davidhart82:


davidhart82:
This should fix #881

nvbn:
Looks good, thanks!

PR #1033, opened 2020-01-05
#N/A: Overwrite links when upgrading Python on MacOS on Travis

scorphus:


scorphus:
Oh. Already obsolete.

Issue #1030, opened 2020-01-02
Add git lfs support


Shayan-To:
Add support for correcting commands like `git lfs checkuot` to `git lfs checkout`.

From their man pages, it has these commands:

- env
- checkout
- clone
- fetch
- fsck
- install
- lock
- locks
- logs
- ls-files
- migrate
- prune
- pull
- push
- status
- track
- uninstall
- unlock
- untrack
- update
- version
- clean
- pointer
- pre-push
- filter-process
- smudge

Link: https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs.1.ronn
(Note: git lfs clone was recently removed.)

cjmartian:
@scorphus this can be closed

scorphus:
Thanks for your contributions, @cjmartian! 🙌

PR #1056, opened 2020-02-24
Lfs support

cjmartian:
This is related to https://github.com/nvbn/thefuck/issues/1030

### DIFF
--- a/tests/rules/test_git_lfs_mistype.py
+++ b/tests/rules/test_git_lfs_mistype.py
@@ -1,0 +1,62 @@
+from thefuck.rules.git_lfs_mistype import match, get_new_command
+from thefuck.types import Command
+
+
+def test_match():
+    err_response = """
+Error: unknown command "evn" for "git-lfs"
+
+Did you mean this?
+	env
+	ext
+
+Run 'git-lfs --help' for usage.
+    """
+    assert match(Command('git lfs evn', err_response))
+
+
+def test_not_match():
+    err_response = 'bash: git: command not found'
+    assert not match(Command('git lfs env', err_response))
+
+
+def test_not_git_command():
+    err_response = """
+Error: unknown command "evn" for "git-lfs"
+
+Did you mean this?
+	env
+	ext
+
+Run 'git-lfs --help' for usage.
+    """
+    assert not match(Command('docker lfs env', err_response))
+
+
+def test_get_new_command():
+    err_response = """
+Error: unknown command "evn" for "git-lfs"
+
+Did you mean this?
+	env
+	ext
+
+Run 'git-lfs --help' for usage.
+    """
+    result = get_new_command(Command('git lfs evn', err_response))
+    expected = 'git lfs env'
+    assert result == expected
+
+
+def test_get_another_new_command():
+    err_response = """
+Error: unknown command "chekout" for "git-lfs"
+
+Did you mean this?
+	checkout
+
+Run 'git-lfs --help' for usage.
+    """
+    result = get_new_command(Command('git lfs chekout', err_response))
+    expected = 'git lfs checkout'
+    assert result == expected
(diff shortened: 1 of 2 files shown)


tests/rules/test_git_lfs_mistype.py
```diff
@@ -0,0 +1,62 @@
...
+Error: unknown command "chekout" for "git-lfs"
+
+Did you mean this?
+        checkout
+
+Run 'git-lfs --help' for usage.
+    """
+    result = get_new_command(Command('git lfs chekout', err_response))
+    expected = 'git lfs checkout'
+    assert result == expected
```
scorphus: I recommend reading on pytest fixtures and trying to replace the repeated `err_response` definitions with a single fixture. There are examples in other test files.

thefuck/rules/git_lfs_mistype.py
```diff
@@ -0,0 +1,18 @@
...
+    '''
+    return 'lfs' in command.script and 'Did you mean this?' in command.output
+
+
+@git_support
+def get_new_command(command):
+    new = command.script.split(' ')
+    recommended = command.output.split('\n')[4].strip()
+    new[2] = recommended
+    return shell.and_(' '.join(new))
```
scorphus: I suggest you take a look at how `thefuck/rules/git_not_command.py` is implemented, as both are very similar. For instance, check how using a regular expression can make the code more robust. Also, the use of both `get_all_matched_commands` and `replace_command` can make the code more flexible and assertive.

tests/rules/test_git_lfs_mistype.py
cjmartian: Thanks for pointing that out, I made that change in my latest commit!

thefuck/rules/git_lfs_mistype.py
cjmartian: That's very helpful and a much better solution. Thank you for pointing that out for me!

thefuck/rules/git_lfs_mistype.py
```diff
@@ -0,0 +1,18 @@
...
+    '''
+    Match a mistyped command
+    '''
+    return 'lfs' in command.script and 'Did you mean this?' in command.output
+
+
+@git_support
+def get_new_command(command):
+    broken_cmd = re.findall(r'Error: unknown command "([^"]*)" for "git-lfs"', command.output)[0]
+    matched = get_all_matched_commands(command.output, ['Did you mean'])
```
scorphus: If you place a breakpoint right after line 17 (e.g `__import__("ipdb").set_trace()` after installing ipdb with `pip install ipdb`) you can see that `matched` contains some undesired content:

```
ipdb> l
     16 def get_new_command(command):
     17     broken_cmd = re.findall(r'Error: unknown command "([^"]+)" for "git-lfs"',
     18                             command.output)[0]
     19     matched = get_all_matched_commands(command.output, 'Did you mean')
     20     __import__("ipdb").set_trace()  # FIXME: BREAKPOINT!!!
---> 21     return replace_command(command, broken_cmd, matched)

ipdb> matched
['env', 'ext', "Run 'git-lfs --help' for usage."]
```

To get the debugger taking input, I ran TheFuck directly, instead of via `fuck`, like this: `thefuck git lfs evn`

Checking `get_all_matched_commands` source code:

https://github.com/nvbn/thefuck/blob/3c542a5b8cde085126fd9197b1ca6967c7adab1e/thefuck/utils.py#L148-L159

You can see that it supports `separator` being a list of strings. So, how about adding `' for usage.'` to the list of separators passed to `get_all_matched_commands`? This way, `matched` will be equal to `['env', 'ext']`.

cjmartian: Thank you for catching that! I didn't even think about looking at that.

I've made the requested change and it no longer contains the unwanted string.

cjmartian:
@scorphus Anytime, I'm happy to help! Thank you for the suggestions and for reviewing my PR!

PR #1063, opened 2020-03-01
[WIP] -Draft PR: Edit commands using the default editor

cjoshmartin:
This PR to implement a feature that would allow you to edit the suggested command in the editor that is set in $EDITOR for terminal other than fish.

Fish natively support line editing, so opening an editor is not need

### DIFF
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -51,8 +51,7 @@ class Fish(Generic):
     def app_alias(self, alias_name):
         if settings.alter_history:
             alter_history = ('    builtin history delete --exact'
-                             ' --case-sensitive -- $fucked_up_command\n'
-                             '    builtin history merge ^ /dev/null\n')
+                             ' --case-sensitive -- $fucked_up_command')
         else:
             alter_history = ''
         # It is VERY important to have the variables declared WITHIN the alias
@@ -61,7 +60,8 @@ class Fish(Generic):
                 '  env TF_SHELL=fish TF_ALIAS={0} PYTHONIOENCODING=utf-8'
                 ' thefuck $fucked_up_command {2} $argv | read -l unfucked_command\n'
                 '  if [ "$unfucked_command" != "" ]\n'
-                '    eval $unfucked_command\n{1}'
+                '    commandline $unfucked_command\n'
+                '    commandline -f execute\n{1}'
                 '  end\n'
                 'end').format(alias_name, alter_history, ARGUMENT_PLACEHOLDER)
 
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -127,3 +127,10 @@ class Fish(Generic):
                     history.write(entry.encode('utf-8'))
                 else:
                     history.write(entry)
+
+    def can_edit(self):
+        return True
+
+    def edit_command(self, command):
+        """Return the shell editable command"""
+        return 'commandline -r "' + command + '"'
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -51,7 +51,7 @@ class Fish(Generic):
     def app_alias(self, alias_name):
         if settings.alter_history:
             alter_history = ('    builtin history delete --exact'
-                             ' --case-sensitive -- $fucked_up_command')
+                             ' --case-sensitive -- $fucked_up_command\n')
         else:
             alter_history = ''
         # It is VERY important to have the variables declared WITHIN the alias
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -60,8 +60,7 @@ class Fish(Generic):
                 '  env TF_SHELL=fish TF_ALIAS={0} PYTHONIOENCODING=utf-8'
                 ' thefuck $fucked_up_command {2} $argv | read -l unfucked_command\n'
                 '  if [ "$unfucked_command" != "" ]\n'
-                '    commandline $unfucked_command\n'
-                '    commandline -f execute\n{1}'
+                '    eval $unfucked_command\n{1}'
                 '  end\n'
                 'end').format(alias_name, alter_history, ARGUMENT_PLACEHOLDER)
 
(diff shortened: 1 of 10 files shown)


thefuck/shells/generic.py
```diff
@@ -121,34 +127,109 @@ def put_to_history(self, command):
...
         """
 
+    def can_edit(self):
+        return False
+
+    def edit_command(self, command):
+        """Spawn default editor (or `vi` if not set) and edit command in a buffer"""
+        # Create a temporary file and write some default text
+        file_path = "The Fuck: Command Edit"
+        file_handle = open(file_path, "w")
```
davidhart82:
```python
with open(file_path, 'w') as file_handle:
```
?

thefuck/shells/generic.py
```diff
@@ -121,34 +127,109 @@ def put_to_history(self, command):
...
+        file_handle.write(command)
+        file_handle.close()
+
+        editor = os.getenv("EDITOR")
```
davidhart82:
```python
editor = os.getenv("EDITOR", "vi")
```
?

thefuck/shells/generic.py
```diff
@@ -121,34 +127,109 @@ def put_to_history(self, command):
...
+        if editor is None:
+            editor = "vi"
+
+        os.system("{} '{}' >/dev/tty".format(editor, file_path))
+
+        file_handle = open(file_path, "r")
```
davidhart82: same as above

thefuck/shells/fish.py
```diff
@@ -127,3 +127,10 @@ def _put_to_history(self, command_script):
                     history.write(entry.encode('utf-8'))
                 else:
                     history.write(entry)
+
+    def can_edit(self):
+        return True
+
+    def edit_command(self, command):
+        """Return the shell editable command"""
+        return 'commandline -r "' + command + '"'
```
davidhart82: we use string formatting rather than concatenation everywhere else

thefuck/shells/generic.py
```diff
@@ -121,34 +127,109 @@ def put_to_history(self, command):
...
+            "unset",
+            "until",
+            "wait",
+            "while",
+        ]
 
     def _get_version(self):
         """Returns the version of the current shell"""
-        return ''
+        return ""
```
davidhart82: these set of changes seem unnecessary?

thefuck/shells/generic.py
```diff
@@ -36,18 +37,20 @@ def to_shell(self, command_script):
...
         warn("Instant mode not supported by your shell")
         return self.app_alias(alias_name)
 
     def _get_history_file_name(self):
-        return ''
+        return ""
 
     def _get_history_line(self, command_script):
-        return ''
+        return ""
```
davidhart82: there does seem to be a fair few superfluous changes?

thefuck/shells/generic.py
cjoshmartin: I did not know this was a feature of `getenv`

thefuck/shells/generic.py
cjoshmartin: I was advised by @scorphus to run [black](https://github.com/psf/black), which is what made these superfluous linting changes.

thefuck/shells/fish.py
cjoshmartin: What do you think about f-strings vs using `.format`?

thefuck/shells/generic.py
davidhart82: OK, sure, no worries.  When Im at work, I'd try and get my team to keep changes (especially inconsequential changes) to a minimum, so it's obvious what files really need reviewing.

thefuck/shells/fish.py
davidhart82: When were f-strings introduced?  Are we still supporting older versions of python3 or even python2 still?

davidhart82: The CI system is still running on older python versions, so you can't use f-strings im afraid!

davidhart82: https://travis-ci.org/github/nvbn/thefuck/builds/661796234?utm_source=github_status&utm_medium=notification

thefuck/shells/generic.py
scorphus: I agree with @davidhart82. It's a good practice to submit only changes that are directly related to the issue/feature they address.

On one hand, Black is great because it leaves no place for discussions about how to format code – out of a sudden, you stop wasting time with everything related to code formatting. On the other hand, when it comes to a codebase that's not previously formatted by it, that generates undesirable noise, such as this very one.

So it's up to the developer to include only the relevant changes.

We can discuss whether or not we should reformat the entire codebase at once and make Black part of the development cycle on another issue.

thefuck/shells/fish.py
scorphus: Oops, that slipped through amidst the various experimentations I was doing. Sorry!

scorphus:
@davidhart82, thank you for all your valuable inputs so far! 👍

cjoshmartin:
A resource I have been looking into to generate the temp file: https://docs.python.org/3/library/tempfile.html

Hopeful, will have this in place this coming weekend or next weekend.

cjoshmartin:
So I was able to switch to using tempfiles (as I said in my last comment) but I can not remember how to run the fuck locally.

scorphus:
Hey, @djh82 👋 I hope all is good! I'd really love to hear your inputs not only regarding this but other issues/PRs here on TheFuck. I've been away for quite a while. Any chances we hear from you? Thanks!

thefuck/shells/generic.py
```diff
@@ -121,6 +122,33 @@ def put_to_history(self, command):
...
+        return False
+
+    def edit_command(self, command):
+        """Spawn default editor (or `vi` if not set) and edit command in a buffer"""
+        # Create a temporary file and write some default text
+        # mktemp somewhere
+
+        editor = os.getenv("EDITOR", "vi")
+
+        tf = tempfile.NamedTemporaryFile(
```
djh82, line 135: NamedTemporaryFile is a context manager, so you can use it with `with`, meaning you don't have to explicitly call close or unlink it.

PR #1069, opened 2020-03-24
Command status code

davidhart82:


davidhart82:
I closed the previous PR, since for some reason it showed an awful lot more commits.  I've added some functional tests as requested.

Not sure what the deal is with PowerShell?

davidhart82:
Still working on this.

davidhart82:
This really doesn't work with the experimental instant mode; that makes it very difficult.

I essentially have 2 options:

1. Return the status from the output_readers (my preferred solution; though there is no return code from read_log.py)
2. Try and get the status from the shell, using whichever magic they provide to do so.  Instant mode kind of gets in the way too here, since it wraps the alias with more commands which end up masking the real status code I need.

PR #1072, opened 2020-03-29
#960: Improve pacman_invalid_option rule

scorphus:
The rule now matches all possible lower case options and only those, not failing for those that cannot be fixed.


nvbn:
Looks good, thanks!

PR #1082, opened 2020-04-28
added gcloud cli

ronandoolan2:
added gcloud cli it was very similar to the existing aws cli

### DIFF
--- a/tests/rules/test_gcloud_cli.py
+++ b/tests/rules/test_gcloud_cli.py
@@ -1,0 +1,63 @@
+import pytest
+
+from thefuck.rules.gcloud_cli import match, get_new_command
+from thefuck.types import Command
+
+
+no_suggestions = '''\
+ERROR: (gcloud) Command name argument expected.
+'''
+
+
+misspelled_command = '''\
+ERROR: (gcloud) Invalid choice: 'comute'.
+Usage: gcloud [optional flags] <group | command>
+  group may be           access-context-manager | ai-platform | alpha | app |
+                         asset | auth | beta | bigtable | builds | components |
+                         composer | compute | config | container | dataflow |
+                         dataproc | datastore | debug | deployment-manager |
+                         dns | domains | endpoints | filestore | firebase |
+                         functions | iam | iot | kms | logging | ml |
+                         ml-engine | organizations | projects | pubsub | redis |
+                         resource-manager | scheduler | services | source |
+                         spanner | sql | tasks | topic
+  command may be         docker | feedback | help | info | init | version
+
+For detailed information on this command and its flags, run:
+  gcloud --help
+'''
+
+
+misspelled_subcommand = '''\
+ERROR: (gcloud.compute) Invalid choice: 'instance'.
+Maybe you meant:
+  gcloud compute instance-groups
+  gcloud compute instance-templates
+  gcloud compute instances
+  gcloud compute target-instances
+
+To search the help text of gcloud commands, run:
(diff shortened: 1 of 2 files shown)


tests/rules/test_gcloud_cli.py
```diff
@@ -0,0 +1,63 @@
...
+  gcloud compute target-instances
+
+To search the help text of gcloud commands, run:
+  gcloud help -- SEARCH_TERMS
+'''
+
+
+
+
+@pytest.mark.parametrize('command', [
```
scorphus, line 46: Quoting build [#1932.2](https://travis-ci.org/github/nvbn/thefuck/jobs/680589249#L460) on TravisCI:
> ./tests/rules/test_gcloud_cli.py:46:1: E303 too many blank lines (4)

ronandoolan2: Thanks didn't see that error. I have removed duplicate blank lines

ronandoolan2: I'll try and get it green best I can, thanks for the input

ronandoolan2: Having a little trouble figuring out why the assertions in pytest aren't returning the correct value, I haven't used pytest before

tests/rules/test_gcloud_cli.py
```diff
@@ -0,0 +1,58 @@
...
+                         dns | domains | endpoints | filestore | firebase |
+                         functions | iam | iot | kms | logging | ml |
+                         ml-engine | organizations | projects | pubsub | redis |
+                         resource-manager | scheduler | services | source |
+                         spanner | sql | tasks | topic
+  command may be         docker | feedback | help | info | init | version
+
+For detailed information on this command and its flags, run:
+  gcloud --help
+'''
```
scorphus, line 26: `misspelled_command` is not used in any test case. Also, `match` does not consider that type of output. You can use other rules as a reference – such as `docker_not_command` – which uses string matching and splitting. Other rules – like `react_native_command_unrecognized` – use regular expression. I'd recommend going for string matching and splitting. What do you think?

tests/rules/test_gcloud_cli.py
scorphus: You got it all green! Awesome! pytest is great and I hope you're learning a good deal of it with this PR 🙂 When in doubt, don't hesitate to ask. Also, tests for other rules may serve as a good reference.

tests/rules/test_gcloud_cli.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.mark.parametrize('command', [
+    Command('gcloud comute instances list', misspelled_subcommand),
+    Command('gcloud compute instance list', misspelled_subcommand)])
+def test_match(command):
+    assert match(command)
+
+
+def test_not_match():
+    assert not match(Command('aws dynamodb invalid', no_suggestions))
```
scorphus, line 49: How about parametrizing this test as well. That way, we have a chance to test a couple of cases more:
* `gcloud` with `no_suggestions`
* `gcloud` with no error output

tests/rules/test_gcloud_cli.py
ronandoolan2: Yeap ill look into those examples and make changes

tests/rules/test_gcloud_cli.py
ronandoolan2: Yeap will do

PR #1095, opened 2020-05-14
Reset colour for Powershell

StormPooper:
This was originally fixed in #805 but appears to have been reverted at some point.

### DIFF
--- a/thefuck/shells/powershell.py
+++ b/thefuck/shells/powershell.py
@@ -12,6 +12,7 @@ class Powershell(Generic):
                '            else { iex "$fuck"; }\n' \
                '        }\n' \
                '    }\n' \
+               '    [Console]::ResetColor() \n' \
                '}\n'
 
     def and_(self, *commands):


scorphus:
Currently, [thefuck/shells/powershell.py](https://github.com/nvbn/thefuck/blob/master/thefuck/shells/powershell.py#L9-L20) contains that line:

https://github.com/nvbn/thefuck/blob/6975d30818792f1b37de702fc93c66023c4c50d5/thefuck/shells/powershell.py#L9-L20

scorphus:
Your branch is branched off of 33a8750, which is approx. 2 years old. If you rebase your commit on top of current master, you'll see it gets skipped by Git.

StormPooper:
Whoops, not sure how I managed that, mustn't have switched back to master when I was looking into it - I'm still having the issue it describes where after selecting a choice it keeps the last colour used (red) until the output changes colours. I'll dig into it and raise an issue or PR, thanks.

PR #1102, opened 2020-06-09
add support for docker daemon not running

soraxas:
For patterns like the following:
```sh
$ docker run bash sleep 100
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.

$ fuck
systemctl start docker; and docker run bash sleep 100 [enter/↑/↓/ctrl+c]
```


djh82:
Is systemctl O/S specific? i.e. does this work on Windows as well as mac? 

Regardless, there are no new tests for this rule and the readme also hasn't been updated.

PR #1103, opened 2020-06-11
[Add] uninstall to rm

sdasasqkim:
Wrong command : uninstall filename
After typing fuck
To : rm filename

### DIFF
--- a/thefuck/rules/rm.py
+++ b/thefuck/rules/rm.py
@@ -1,0 +1,11 @@
+import os
+from thefuck.utils import for_app
+
+@for_app('uninstall', at_least=1)
+def match(command):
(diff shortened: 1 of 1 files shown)


thefuck/rules/rm.py
```diff
@@ -0,0 +1,11 @@
+import os
+from thefuck.utils import for_app
+
+@for_app('uninstall', at_least=1)
+def match(command):
+	return (
+		command.output.startswith('uninstall: ')and
+		'command not found' incommand output
```
djh82, line 8: This doesn't look like valid syntax?

djh82:
Be good to have a test for this rule!

PR #1104, opened 2020-06-13
Allow editing auto-correction by fuck

soraxas:


thefuck/ui.py
```diff
@@ -84,6 +87,20 @@ def select_command(corrected_commands):
         if action == const.ACTION_SELECT:
             sys.stderr.write('\n')
             return selector.value
+        elif action == const.ACTION_TO_BUFFER:
+            sys.stderr.write('\n')
+            # use a perl trick to pipe selection to terminal input buffer
```
djh82, line 92: Does this work on all shells? Thinking particularly powershell for windows. This might want to be a per shell enabled feature?

thefuck/ui.py
```diff
@@ -20,6 +21,8 @@ def read_actions():
             yield const.ACTION_NEXT
         elif key in (const.KEY_CTRL_C, 'q'):
             yield const.ACTION_ABORT
+        elif key in ('\t'):
```
djh82, line 24: This hasn't been added to the prompt, so the user doesn't know this is an option?

StefanParenti:
Hello, was just wondering what happened to this pull request? Why was it not merged?

PR #1106, opened 2020-06-15
Add rule to remove a doubled-up "git clone" in a git clone command.

nyanotech:
Some git hosts will copy the entire clone command, while others just copy the url, so typing "git clone ", then pasting a git url that already has a "git clone " on the front is a somewhat common issue.

### DIFF
--- a/tests/rules/test_git_clone_git_clone.py
+++ b/tests/rules/test_git_clone_git_clone.py
@@ -1,0 +1,24 @@
+from thefuck.rules.git_clone_git_clone import match, get_new_command
+from thefuck.types import Command
+
+
+output_clean = """
+fatal: Too many arguments.
+
+usage: git clone [<options>] [--] <repo> [<dir>]
+"""
+
+
+def test_match():
+    assert match(Command('git clone git clone foo', output_clean))
+
+
+def test_not_match():
+    assert not match(Command('', ''))
+    assert not match(Command('git branch', ''))
+    assert not match(Command('git clone foo', ''))
+    assert not match(Command('git clone foo bar baz', output_clean))
+
+
+def test_get_new_command():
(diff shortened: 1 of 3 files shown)


tests/rules/test_git_clone_git_clone.py
```diff
@@ -0,0 +1,61 @@
...
+    -c, --config <key=value>
+                          set config inside the new repository
+    --server-option <server-specific>
+                          option to transmit
+    -4, --ipv4            use IPv4 addresses only
+    -6, --ipv6            use IPv6 addresses only
+    --filter <args>       object filtering
+    --remote-submodules   any cloned submodules will use their remote-tracking branch
+    --sparse              initialize sparse-checkout file to include only files at root
+"""
```
scorphus, line 53: Nit: The output doesn't actually need to be this long. How about limiting it to the first three lines? As in:
```python
output_clean = """fatal: Too many arguments.

usage: git clone [<options>] [--] <repo> [<dir>]

(snip)
"""
```

tests/rules/test_git_clone_git_clone.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_match():
+    assert match(Command('git clone git clone foo', output_clean))
+
```
scorphus, line 58: How about adding a `test_not_match` case? e.g.: https://github.com/nvbn/thefuck/blob/365db1ee41d7bcd9f62d910f3b20361ea3499d92/tests/rules/test_git_branch_list.py#L10-L14

nyanotech: Alright, I think this should catch the likely mis-matches that could happen.

nvbn:
Looks good, thanks!

PR #1114, opened 2020-07-16
* Pamac: Correct invalid operation

earboxer:
Adds a rule to make incorrectly typed pamac commands correct (like how incorrectly typed apt commands work).


scorphus (changes requested):
Thanks for contributing, @earboxer! Would you be willing to fix the issues raised by the checks? Such as add tests and a description to README.md

PR #1119, opened 2020-07-20
Fix quotes problem with shell.split_command

SahandAslani:


kevkevinpal:
I'm trying to get an idea of what the issue is fixing is there a issue number or an example that you have?

PR #1121, opened 2020-08-02
Import CommandNotFound system package on Debian

SuperSandro2000:
Debian does not have a python{,3}-commandnotfound package and the command-not-found package places the python module in another directory.
To import it correctly append the path to the sys.path


thefuck/rules/apt_get.py
```diff
@@ -1,8 +1,15 @@
...
+
 from types import ModuleType
 from thefuck.specific.apt import apt_available
 from thefuck.utils import memoize, which
 from thefuck.shells import shell
 
+# Debian does not have a python{,3}-commandnotfound package
+# and the command-not-found package places the python module in another directory
+# To import it correctly append the path to the sys.path
+sys.path.append('/usr/share/command-not-found/')
```
djh82, line 11: Although a trivial change, I'm not convinced that adding a hardcoded path to the sys path, by default, is the the most appropriate way of fixing this.

SuperSandro2000: ``sys.path`` contains by default ``/usr/lib/python39.zip`` which also does not exist. This can't break anything and gating it behind an if just makes it slower.

SuperSandro2000:
bump, this is still an issue and still the easiest fix

PR #1152, opened 2020-12-27
Fix docker_login rule

divykj:
`docker login` rule used `&&` directly in the output, which could fail in older PowerShell and fish shell versions.

Replaced it with thefuck's `shell._and` helper

### DIFF
--- a/thefuck/rules/docker_login.py
+++ b/thefuck/rules/docker_login.py
@@ -1,4 +1,5 @@
 from thefuck.utils import for_app
+from thefuck.shells import shell
 
 
 @for_app('docker')
@@ -9,4 +10,4 @@ def match(command):
 
 
 def get_new_command(command):
-    return 'docker login && {}'.format(command.script)
+    return shell.and_('docker login', command.script)


tests/rules/test_docker_login.py
```diff
@@ -1,5 +1,6 @@
 from thefuck.rules.docker_login import match, get_new_command
 from thefuck.types import Command
+from thefuck.shells import shell
```
scorphus, line 3: No need to use `and_` in tests, as the [shell is set to `Generic`](https://github.com/nvbn/thefuck/blob/7f97818663de9351ac7e2c6314ed94184811d62a/tests/conftest.py#L7).

```suggestion
```

tests/rules/test_docker_login.py
```diff
@@ -33,5 +34,5 @@ def test_match():
 
 
 def test_get_new_command():
-    assert get_new_command(Command('docker build -t artifactory:9090/foo/bar:fdb7c6d .', '')) == 'docker login && docker build -t artifactory:9090/foo/bar:fdb7c6d .'
-    assert get_new_command(Command('docker push artifactory:9090/foo/bar:fdb7c6d', '')) == 'docker login && docker push artifactory:9090/foo/bar:fdb7c6d'
+    assert get_new_command(Command('docker build -t artifactory:9090/foo/bar:fdb7c6d .', '')) == shell.and_('docker login', 'docker build -t artifactory:9090/foo/bar:fdb7c6d .')
+    assert get_new_command(Command('docker push artifactory:9090/foo/bar:fdb7c6d', '')) == shell.and_('docker login', 'docker push artifactory:9090/foo/bar:fdb7c6d')
```
scorphus, line 38: Ditto.
```suggestion
    assert get_new_command(Command('docker build -t artifactory:9090/foo/bar:fdb7c6d .', '')) == 'docker login && docker build -t artifactory:9090/foo/bar:fdb7c6d .'
    assert get_new_command(Command('docker push artifactory:9090/foo/bar:fdb7c6d', '')) == 'docker login && docker push artifactory:9090/foo/bar:fdb7c6d'
```

scorphus (changes requested):
Sorry for the delay, @divykj. Please consider my suggestions below.

scorphus (approved):

scorphus:
Thanks, @divykj.

Issue #1144, opened 2020-11-09
correct 'cask reinstall' to 'brew cask reinstall' instead of 'brew cask install'


grafst:
The Fuck 3.29 using Python 3.7.4 and ZSH 5.7.1
MacOS 10.15.6

Hi, I hope this is the right place to suggest this.

divykj:
Hey, I am not able to find the rule that is correcting 'cask reinstall' to 'brew cask install'.
Could you please provide the output of `THEFUCK_DEBUG=true thefuck cask reinstall`.


> Hi, I hope this is the right place to suggest this.

Yes, this is absolutely the right place!

scorphus:
This is quite probably the `history` rule kicking in. There's probably one entry such as `brew cask install foo` in your history that is a close-enough match to, for example, `cask install bar`. Unfucking this is beyond the capabilities of any existing rule.

But that can change with a new rule that looks for commands starting with `cask` and then prepend `brew` to it.

grafst:
@scorphus that makes sense
thanks for explaining
maybe some time i find a chance to look into creating that rule

Armaxxx:
I'll attempt to create the rule :)

PR #1155, opened 2021-01-05
Added brew cask reinstall to the posibilities

Armaxxx:
Can somebody help me check this? I am not able to test it my Mac just died.
It attempts to solve  #1144 by just adding the option of brew cask reinstall along with brew install and brew cask install, since I couldn't manage to create a new rule.

### DIFF
--- a/thefuck/rules/brew_cask_dependency.py
+++ b/thefuck/rules/brew_cask_dependency.py
@@ -6,7 +6,8 @@ from thefuck.specific.brew import brew_available
 @for_app('brew')
 def match(command):
     return (u'install' in command.script_parts
-            and u'brew cask install' in command.output)
+            and u'brew cask install' in command.output
+            and u'brew cask reinstall' in command.output)
 
 
 @eager


thefuck/rules/brew_cask_dependency.py
```diff
@@ -6,7 +6,8 @@
 @for_app('brew')
 def match(command):
     return (u'install' in command.script_parts
-            and u'brew cask install' in command.output)
+            and u'brew cask install' in command.output
+            and u'brew cask reinstall' in command.output)
```
scorphus, line 10: From what I understand, that should be an `or` rather than an `and`.

scorphus (commented):
Anyway, since [2.6.0](https://brew.sh/2020/12/01/homebrew-2.6.0/) release, Homebrew has deprecated `cask` commands. So, I wonder what we do with this rule.

thefuck/rules/brew_cask_dependency.py
djh82: An updated test would certainly help here!

PR #1160, opened 2021-01-20
Add cache code snippet

equwal:
I stopped using thefuck because it added over 1.0s on my machine's shell
startup. A cache removes this issue.

This is especially a problem if you want multiple aliases.


equwal:
Made a wiki page instead.

https://github.com/nvbn/thefuck/wiki/Cache-aliases-Startup

PR #1162, opened 2021-01-25
Allow multiple alias arguments

equwal:
This change makes up somewhat for thefuck's >100ms time to print the
aliases.

Some profiling shows that 50% of that time is spent loading a bunch of
unsued packages, and the other 50% is unavoidable overhead for starting
a python script.

These are valid:
thefuck --alias f k fuck fk # new!
thefuck --alias fk # one arg form
thefuck --alias # uses fuck by default still (backward compatibility)
thefuck --alias fk --enable-experimental-instant-mode # also supported!

Before, setting 10 aliases meant a 1second shell startup. Now:

$ time thefuck --alias a b c d e f g h i j
thefuck --alias a b c d e f g h i j  0.20s user 0.04s system 96% cpu 0.256 total

Not bad!

### DIFF
--- a/thefuck/argument_parser.py
+++ b/thefuck/argument_parser.py
@@ -1,7 +1,27 @@
 import sys
-from argparse import ArgumentParser, SUPPRESS
+from argparse import ArgumentParser, SUPPRESS, Action
 from .const import ARGUMENT_PLACEHOLDER
 from .utils import get_alias
+import os
+
+class Alias(Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+
+        if option_string:
+            if not values:
+                setattr(namespace, 'alias', [os.environ.get('TF_alias', 'fuck')])
+
+        if hasattr(namespace, 'alias'):
+            current_values = getattr(namespace, 'alias')
+            try:
+                current_values.extend(values)
+            except AttributeError:
+                current_values = values
+            finally:
+                setattr(namespace, 'alias', current_values)
+        else:
+            setattr(namespace, 'alias', values)
+
 
 
 class Parser(object):
@@ -22,8 +42,8 @@ class Parser(object):
             help="show program's version number and exit")
         self._parser.add_argument(
             '-a', '--alias',
-            nargs='?',
-            const=get_alias(),
+            nargs='*',
+            action=Alias,
             help='[custom-alias-name] prints alias for current shell')
         self._parser.add_argument(
             '-l', '--shell-logger',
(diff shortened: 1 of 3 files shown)


equwal:
Bug in my PR:

TF_ALIAS doesn't get set, even though it is in the function:
```bash
function x () {
    TF_PYTHONIOENCODING=$PYTHONIOENCODING;
    export TF_SHELL=bash;
    export TF_ALIAS=x;                <---------------------
    export TF_SHELL_ALIASES=$(alias);
    export TF_HISTORY=$(fc -ln -10);
    export PYTHONIOENCODING=utf-8;
    TF_CMD=$(
        thefuck THEFUCK_ARGUMENT_PLACEHOLDER "$@"
    ) && eval "$TF_CMD";
    unset TF_HISTORY;
    export PYTHONIOENCODING=$TF_PYTHONIOENCODING;
    history -s $TF_CMD;
}
```

equwal:
The problem is a littlle bit deeper than that, since really it doesn't make sense to have this variable that has "the alias". Of course, you can have more than one alias!

So the variable should just be done away with, and the tests for it.

equwal:
Furthermore, the "feature" of printing out the current alias magically with a nullary `--alias` also makes no sense when using multiple aliases, and has to go.

scorphus:
Thanks for taking the time to contribute, that's appreciated.

I'm afraid this feature won't benefit the majority of users. Even if it would — and it certainly does benefit some users — it can be solved in the shell by using `alias`. Every change adds complexity to the code. Code that need to be maintained. The less code we need to add the better.

Could you please try using `alias` and see if it is quick enough?

I tried on my end and the difference wasn't noticeable. Before changing `.bashrc` to include additional aliases:

```bash
$ cat .bashrc
eval "$(thefuck --alias)"
$ time bash .bashrc

real	0m0.163s
user	0m0.088s
sys	0m0.071s
```

After changing `.bashrc` to include additional aliases:

```bash
$ cat .bashrc
eval "$(thefuck --alias)"
alias f=fuck
alias k=fuck
alias fk=fuck
alias fu=fuck
alias fck=fuck
alias fuc=fuck
alias fuk=fuck
$ time bash .bashrc

real	0m0.165s
user	0m0.088s
sys	0m0.073s
```

equwal:
> Could you please try using `alias` and see if it is quick enough?

> $ cat .bashrc
> eval "$(thefuck --alias)"
> alias f=fuck
> alias k=fuck
> [...]

Oops, that really does negate the need for this issue indeed; I wasn't aware that this was possible!

Still, I think thefuck is way too slow at 165ms  just to make an alias. Would be good to pull that code out of all of the packaging systems and whatnot, could probably get it down to 50ms without any change in functionality.

PR #1167, opened 2021-02-14
Introducing the cd-cs feature

s1ddly:
I found in my usage that due to the proximity of the 's' and 'd' key, I would commonly type cs when I mean to say cd.
I am hoping to address that with this update to the rules.

### DIFF
--- a/thefuck/rules/cd_cs.py
+++ b/thefuck/rules/cd_cs.py
@@ -1,7 +1,10 @@
 # Redirects cs to cd when there is a typo
-# Due to the proximity of the keys - d and s, I found this to be a common problem for me
-# > cs /etc/
+# Due to the proximity of the keys — d and s — this seems like a common typo
+# ~ > cs /etc/
 # cs: command not found
+# ~ > fuck
+# cd /etc/ [enter/↑/↓/ctrl+c]
+# /etc >
 
 
 def match(command):
--- a/thefuck/rules/cd_cs.py
+++ b/thefuck/rules/cd_cs.py
@@ -1,5 +1,5 @@
 # Redirects cs to cd when there is a typo
-# Due to the proximity of the keys — d and s — this seems like a common typo
+# Due to the proximity of the keys - d and s - this seems like a common typo
 # ~ > cs /etc/
 # cs: command not found
 # ~ > fuck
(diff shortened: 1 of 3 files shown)


s1ddly:
@scorphus have you had a chance to look at this yet?

s1ddly:
FB_IMG_1616075794372

BimboPolitics:
Following this github is the best decision ive made all week

thefuck/rules/cd_cs.py
```diff
@@ -0,0 +1,16 @@
+# Redirects cs to cd when there is a typo
+# Due to the proximity of the keys - d and s, I found this to be a common problem for me
+# > cs /etc/
+# cs: command not found
```
scorphus, line 4: Nit: can I suggest a more informative head comment?
```suggestion
# Redirects cs to cd when there is a typo
# Due to the proximity of the keys — d and s — this seems like a common typo
# ~ > cs /etc/
# cs: command not found
# ~ > fuck
# cd /etc/ [enter/↑/↓/ctrl+c]
# /etc >
```

scorphus (approved):
Hey, @s1ddly 👋 🤣

Sorry for the delay! I was wondering if we could fix this [some other way](https://github.com/scorphus/thefuck/tree/typo-distance) — that being the outcome of a hackthon I mentored a couple weeks ago. But it looks like that will need a bit more working.

Thank you for your contribution!

s1ddly:
Hey Mate,

I found that some of the python 2.7 build test cases failed on this one.
It seems that it doesn't like the up and down arrows in the comment.
Are we good to proceed with the failed tests or should I remove the offending characters?

Thanks, Sid.

thefuck/rules/cd_cs.py
```diff
@@ -0,0 +1,19 @@
+# Redirects cs to cd when there is a typo
```
scorphus, line 1: This should fix it:

```suggestion
# -*- encoding: utf-8 -*-

# Redirects cs to cd when there is a typo
```

scorphus (changes requested):

s1ddly:
Looks like that fixed it.
Thats a pretty useful trick, I'll be sure to remember that for my future projects :)

scorphus (approved):

scorphus:
Thanks, @s1ddly!

PR #1186, opened 2021-04-13
Cd rules

ICalhoun:
I fixed cd_mkdir.py to work with powershell in linux and windows. I also modified it and added it as chdir.py so that the same corrections can be made for the chdir command, which is the cd command, in case anyone uses it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ following rules are enabled by default:
 * `cd_cs` &ndash; changes `cs` to `cd`;
 * `cd_mkdir` &ndash; creates directories before cd'ing into them;
 * `cd_parent` &ndash; changes `cd..` to `cd ..`;
+* `chdir` &ndash; creates directories before changing into them using chdir;
 * `chmod_x` &ndash; add execution bit;
 * `choco_install` &ndash; append common suffixes for chocolatey packages;
 * `composer_not_command` &ndash; fixes composer command name;
--- a/thefuck/rules/cd_mkdir.py
+++ b/thefuck/rules/cd_mkdir.py
@@ -11,7 +11,9 @@ def match(command):
         command.script.startswith('cd ') and any((
             'no such file or directory' in command.output.lower(),
             'cd: can\'t cd to' in command.output.lower(),
-            'does not exist' in command.output.lower()
+            'does not exist' in command.output.lower(),
+            'cannot find the path' in command.output.lower(),
+            'cannot find path' in command.output.lower()
         )))
 
 
--- a/thefuck/rules/chdir.py
+++ b/thefuck/rules/chdir.py
@@ -8,13 +8,14 @@ from thefuck.shells import shell
 @for_app('chdir')
 def match(command):
     return (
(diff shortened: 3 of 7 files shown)


thefuck/rules/cd_mkdir.py
```diff
@@ -11,11 +10,13 @@ def match(command):
         command.script.startswith('cd ') and any((
             'no such file or directory' in command.output.lower(),
             'cd: can\'t cd to' in command.output.lower(),
-            'does not exist' in command.output.lower()
```
djh82, line 14: Wonder whether, now this list is getting longer, something like this would be better:

```python
if any(word in command.output.lower() for word in does_not_exist_list)
```

thefuck/rules/cd_mkdir.py
```diff
@@ -11,11 +10,13 @@ def match(command):
...
+            'does not exist' in command.output.lower(),
+            'cannot find the path' in command.output.lower(),
+            'cannot find path' in command.output.lower()
         )))
 
 
 @sudo_support
 def get_new_command(command):
-    repl = shell.and_('mkdir -p \\1', 'cd \\1')
+    repl = 'mkdir -p \\1; cd \\1'
```
djh82, line 21: Why the change here?

README.md
```diff
@@ -456,6 +457,7 @@ Or via environment variables:
 * `THEFUCK_RULES` &ndash; list of enabled rules, like `DEFAULT_RULES:rm_root` or `sudo:no_command`;
 * `THEFUCK_EXCLUDE_RULES` &ndash; list of disabled rules, like `git_pull:git_push`;
 * `THEFUCK_REQUIRE_CONFIRMATION` &ndash; require confirmation before running new command, `true/false`;
+* `THEFUCK_REQUIRE_DOUBLE_CONFIRMATION` &ndash; require double confirmation before running potentially volitile commands (eg. `reboot`), value of `true/false`;
```
djh82, line 460: This really should have been a separate PR, since it makes it default now both sets of changes have to be approved before either can get merged.

thefuck/rules/chdir.py
```diff
@@ -0,0 +1,26 @@
+import re
```
djh82, line 1: This file is very similar to the cd variant and so we have duplicated logic. Do we think it's possibl/worthwhile combining them?

thefuck/rules/chdir.py
```diff
@@ -0,0 +1,26 @@
...
+            'cannot find path' in command.output.lower(),
+            'can\'t cd to' in command.output.lower()
+        )))
+
+
+@sudo_support
+def get_new_command(command):
+    shell_name = shell.info()
+    if shell_name == 'pwsh' or shell_name == 'powershell':
+        new_command = ('mkdir -p \\1' + '; chdir\\1')
```
djh82, line 22: Similar to previous comment, I think it's prefer to 'and' these.

tpevers16 (approved):

PR #1196, opened 2021-05-05
Avoid using pkg_resources

felixonmars:
pkg_resources is expensive to load, and it's an external module provided
by the setuptools project.

Let's use importlib.metadata from Python 3.8+'s stdlib when available, which
is much faster. This also erases setuptools from thefuck's runtime
dependency.

Simple benchmark data for `thefuck --version`:

Before: 0.602s
After: 0.169s


scorphus (approved):
Thanks, @felixonmars! 👍

PR #1224, opened 2021-08-02
#697: Encode expanded script on Python 2

### LINKED ISSUE #697: Non-ASCII character in last command crashes The Fuck
If the last command contains a non-ASCII character, The Fuck crashes with a UnicodeEncodeError.

Such an error can easily happen on non-English keyboards, e.g. on my Hungarian keyboard, it's very easy to misspell 'ls" as "és", or "ps aux" as "ős aux".

Reproduction:
with The Fuck configured, run "ápt-get moo", then "fuck". 
The following error appears:
Traceback (most recent call last):
  File "/home/osboxes/venv/bin/thefuck", line 11, in <module>
    sys.exit(main())
  File "/home/osboxes/venv/local/lib/python2.7/site-packages/thefuck/main.py", line 53, in main
    fix_command(known_args)
  File "/home/osboxes/venv/local/lib/python2.7/site-packages/thefuck/main.py", line 29, in fix_command
    command = types.Command.from_raw_script(raw_command)
  File "/home/osboxes/venv/local/lib/python2.7/site-packages/thefuck/types.py", line 76, in from_raw_script
    stdout, stderr = get_output(script, expanded)
  File "/home/osboxes/venv/local/lib/python2.7/site-packages/thefuck/output_readers/__init__.py", line 18, in get_output
    return rerun.get_output(script, expanded)
  File "/home/osboxes/venv/local/lib/python2.7/site-packages/thefuck/output_readers/rerun.py", line 42, in get_output
    is_slow = shlex.split(expanded) in settings.slow_commands
  File "/usr/lib/python2.7/shlex.py", line 275, in split
    lex = shlex(s, posix=posix)
  File "/usr/lib/python2.7/shlex.py", line 25, in __init__
    instream = StringIO(instream)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe1' in position 0: ordinal not in range(128)

If you instead type "opt-get moo", then "fuck", it works fine.

scorphus:
Fix #697

### DIFF
--- a/thefuck/output_readers/rerun.py
+++ b/thefuck/output_readers/rerun.py
@@ -1,5 +1,6 @@
 import os
 import shlex
+import six
 from subprocess import Popen, PIPE, STDOUT
 from psutil import AccessDenied, Process, TimeoutExpired
 from .. import logs
@@ -52,6 +53,9 @@ def get_output(script, expanded):
     """
     env = dict(os.environ)
     env.update(settings.env)
+
+    if six.PY2:
+        expanded = expanded.encode('utf-8')
 
     split_expand = shlex.split(expanded)
     is_slow = split_expand[0] in settings.slow_commands if split_expand else False
--- a/tests/output_readers/test_rerun.py
+++ b/tests/output_readers/test_rerun.py
@@ -30,6 +30,11 @@ class TestRerun(object):
         actual = rerun.get_output('', '')
         assert actual == expected
 
+    @patch('thefuck.output_readers.rerun._wait_output')
+    def test_get_output_unicode_misspell(self, wait_output_mock):
+        rerun.get_output(u'pácman', u'pácman')
+        wait_output_mock.assert_called_once()
+
     def test_wait_output_is_slow(self, settings):
         assert rerun._wait_output(Mock(), True)
         self.proc_mock.wait.assert_called_once_with(settings.wait_slow_command)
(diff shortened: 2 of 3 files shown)


thefuck/output_readers/rerun.py
```diff
@@ -1,5 +1,6 @@
 import os
 import shlex
+import six
```
djh82, line 3: Is this a new dependency? Presumably requirements.txt and the like will also need updating?

scorphus:
Hey, @djh82! 👋 

I'd really love if you could drop a review here and/or your thoughts — apart from "you shouldn't be wasting time on Python 2.7, Pablo." hahaha 🙈😁 — I promise that's the last one thou.

Thanks in advance!!

djh82:
Looks reasonable to me! My only question is about the use of replace for the errors argument - there are very few other occasion where that is specified (and they might only be log prints?), the rest of the codebase defaults to strict.

Given we're expecting folk to have typed these commands, we should be expecting valid utf8? I don't see the harm in erroring - I think it would be more surprising to the user to replace characters?

scorphus:
I see your point, @djh82. My thinking was in the case where the user might be in a situation where the encoding issue cannot be solved -- for some reason such as the user is not able to or doesn't know how to.

I've been thinking that thefuck could then issue a warning and do the replacement (or ignore?). That way, it works when something is not right but the user is aware of the problem. What do you think?

djh82:
I think adding a warning is a reasonable compromise. It's always a compromise, but I aim for the least surprising action for the user.

scorphus:
👍 @djh82, thanks to your inputs, while working a bit more around #697, I got to the same conclusion as you did. If for some reason `expanded` cannot be encoded, it's better to just let it explode so the user knows there's something wrong with their setup.

PR #1233, opened 2021-09-05
README: Use `--user` with pip, not `sudo pip`

benjamin-rood:
`sudo` with `pip` is [considered unsafe](https://stackoverflow.com/a/22517157/2469559).

Instead, run such commands with the `--user` argument to get the same result _safely._


scorphus (approved):
Thanks, @benjamin-rood :raised_hands:

Issue #1036, opened 2020-01-19
Slow rule eval on WSL


mxplusb:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

```sh
% thefuck --version
The Fuck 3.29 using Python 3.8.1 and ZSH 5.4.2
```

Your system (Debian 7, ArchLinux, Windows, etc.):

```sh
% uname -r
4.19.84-microsoft-standard
```

![image](https://user-images.githubusercontent.com/7537841/72685523-539df900-3aa8-11ea-81af-5690906ddcad.png)

How to reproduce the bug:

```sh
# linuxbrew using latest.
% brew install -v thefuck
% echo 'eval $(thefuck --alias fuck)' >> ~/.zshrc
% source ~/.zshrc

# some repo
% gitstatus
zsh: command not found: gitstatus

# this takes like 30 seconds.
% fuck
```

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

```sh
 % fuck --debug
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': True,
 'no_colors': False,
 'num_close_matches': 3,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': PosixPath('/c/Users/MikeLloyd/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
[WARN] PS1 doesn't contain user command mark, please ensure that PS1 is not changed after The Fuck alias initialization
DEBUG: Importing rule: adb_unknown_command; took: 0:00:00.000184
DEBUG: Importing rule: ag_literal; took: 0:00:00.000366
DEBUG: Importing rule: apt_get; took: 0:00:00.001400
DEBUG: Importing rule: apt_get_search; took: 0:00:00.000299
DEBUG: Importing rule: apt_invalid_operation; took: 0:00:00.000714
DEBUG: Importing rule: apt_list_upgradable; took: 0:00:00.000370
[...]

PR #1236, opened 2021-09-16
Change readme installation to have WSL section

Temez1:
Add installation to have a WSL section so that WSL users don't need to start debugging the slowness after installation. Now it's stated explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,14 @@ sudo apt install python3-dev python3-pip python3-setuptools
 pip3 install thefuck --user
 ```
 
+On WSL after installing, change the `~/.config/thefuck/settings.py` to include
+
+```
+excluded_search_path_prefixes = ['/mnt/']
+```
+
+to avoid slowness. For more info, see [this comment](https://github.com/nvbn/thefuck/issues/1036#issuecomment-773641630)
+
 On FreeBSD, install *The Fuck* with the following commands:
 ```bash
 pkg install thefuck


README.md
```diff
@@ -127,6 +127,14 @@ sudo apt install python3-dev python3-pip python3-setuptools
 pip3 install thefuck --user
 ```
 
+On WSL after installing, change the `~/.config/thefuck/settings.py` to include
+
+```
+excluded_search_path_prefixes = ['/mnt/']
+```
+
+to avoid slowness. For more info, see [this comment](https://github.com/nvbn/thefuck/issues/1036#issuecomment-773641630)
```
scorphus, line 136:
```suggestion
to avoid slowness. For more info, see [this comment](https://github.com/nvbn/thefuck/issues/1036#issuecomment-773641630).
```

scorphus (changes requested):
Looks good. Please, consider my suggestion below. Also, what do you think about mentioning the wiki? https://github.com/nvbn/thefuck/wiki/Troubleshooting#wsl-fix-slowness

Temez1:
I got the idea from [this](https://github.com/nvbn/thefuck/issues/1036#issuecomment-761173395) comment that wiki solution is suboptimal. I didn't do anything else than this and it works great in WSL. I think that the wiki page should be changed to have this current solution.

scorphus:
> […] I think that the wiki page should be changed to have this current solution.

Could you please change it? It would be great to have that fixed.

Temez1:
I just copypasted it there lol

PR #1243, opened 2021-11-02
Add `ping` rule

mutoo:
**Background:** I often made this mistake by copying a URL from the browser's address bar and pasting it to the `ping` command. And it always yields that error:

```
$ ping https://github.com/
ping: cannot resolve https://github.com/: Unknown host
```

I expected `thefuck` could fix this by doing

```
$ fuck
ping github.com
```

So I created this PR to add the `ping` rule.

### DIFF
--- a/tests/rules/test_ping.py
+++ b/tests/rules/test_ping.py
@@ -1,0 +1,41 @@
+import pytest
+from thefuck.rules.ping import get_new_command, match
+from thefuck.types import Command
+
+
+cannot_resolve_output = '''
+ping: cannot resolve {}: Unknown host
+'''.format
+
+resolve_output = '''
+PING {} ({}): 56 data bytes
+64 bytes from {}: icmp_seq=0 ttl=64 time=0.050 ms
+'''.format
+
+@pytest.mark.parametrize('command', [
+    Command('ping https://google.com', cannot_resolve_output('https://google.com')),
+    Command('ping http://google.com', cannot_resolve_output('http://google.com')),
+    Command('ping http://google.com/', cannot_resolve_output('http://google.com/')),
+    Command('ping http://google.com/?q=test', cannot_resolve_output('http://google.com/?q=test')),
+    Command('ping ftp://192.168.0.1', cannot_resolve_output('ftp://192.168.0.1')),
+    Command('ping -c 10 ftp://192.168.0.1', cannot_resolve_output('ftp://192.168.0.1')), # allow options
+])
+def test_match(command):
+    assert match(command)
+
+
+@pytest.mark.parametrize('command', [
+    Command('ping http:/google.com/', cannot_resolve_output('http:/google.com/')), # skip invalid url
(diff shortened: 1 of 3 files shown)


import-brain (approved):

tests/rules/test_ping.py
```diff
@@ -0,0 +1,41 @@
...
+def test_not_match(command):
+    assert not match(command)
+
+
+@pytest.mark.parametrize('command, new_command', [
+    (Command('ping https://google.com', cannot_resolve_output('https://google.com')), 'ping google.com'),
+    (Command('ping -c 10 https://google.com', cannot_resolve_output('https://google.com')), 'ping -c 10 google.com'),
+])
+def test_get_new_command(command, new_command):
+    assert get_new_command(command) == new_command
```
scorphus, line 41: It's great that you add tests for the rule!

Could you please use fixtures and the briefest possible parameters? Not only it reads better, it's also easier to reason about and, more importntly, maintain in the future. Please have a look at an example:

https://github.com/nvbn/thefuck/blob/4c7479b3adcf8715a93d0c48e1ece83a35cda50d/tests/rules/test_brew_uninstall.py#L1-L31

thefuck/rules/ping.py
```diff
@@ -0,0 +1,19 @@
...
+from six.moves.urllib.parse import urlparse
+from thefuck.utils import for_app
+
+
+@for_app('ping')
+def match(command):
+    # It only fix if the target host is a valid url
+    try:
+        results = urlparse(command.script_parts[-1])
+        is_valid_url = results.hostname is not None
```
scorphus, line 11: Would be nice to extract this to a function, have it memoized, check first if `"cannot resolve"` is part of the output. Then, only if that is the case, call the function, check its result (if the URL is valid) and reuse it in `get_new_command`. Can you please do it?

thefuck/rules/ping.py
```diff
@@ -0,0 +1,19 @@
...
+    except Exception:
+        is_valid_url = False
+    return 'cannot resolve' in command.output and is_valid_url
+
+
+def get_new_command(command):
+    result = urlparse(command.script_parts[-1])
+    return ' '.join(command.script_parts[:-1]) + ' ' + result.hostname
```
scorphus, line 19: What if the user wants to fix `ping https://google.com/ -c 5`?

scorphus (commented):
Thanks for contributing! Please consider my suggestions below.

tests/rules/test_ping.py
mutoo: Thanks for the example. I had refactored the tests with fixtures. It's the first time I used the pytest. I hope I am doing it correctly 😄

thefuck/rules/ping.py
mutoo: I also refactored the way to look up a valid URL among the options. It's should work now.

PR #1248, opened 2021-11-23
Add support for Python 3.10 and 3.11

hugovk:
Python 3.10 was released on 2021-10-04:

image

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955

Drop the dot in `pytest`: 

* https://twitter.com/pytestdotorg/status/753767547866972160

Update min Python 3 version required in `README` to 3.5 to match `setup.py`.

Add `python_requires` to help pip.

PR #1249, opened 2021-11-23
Drop support for EOL Python 3.5

hugovk:
(Includes https://github.com/nvbn/thefuck/pull/1248 to avoid conflicts, the [last commit](e7215062ae65d6ac34e0391404db59bc2b64a8de) here is unique. Happy to rebase or split it out.)

Drop support for EOL Python 3.5:

| cycle | latest |  release   |    eol     |
|:------|:-------|:----------:|:----------:|
| 3.6   | 3.6.15 | 2016-12-23 | 2021-12-23 |
| 3.5   | 3.5.10 | 2015-09-30 | 2020-09-13 |
| 2.7   | 2.7.18 | 2010-07-03 | 2020-01-01 |

It's also little used.

Here's the pip installs for thefuck from PyPI for October 2021:

| category | percent | downloads |
|:---------|--------:|----------:|
| 3.8      |  31.04% |     1,950 |
| 3.9      |  26.97% |     1,694 |
| null     |  25.74% |     1,617 |
| 3.6      |   5.46% |       343 |
| 3.10     |   3.82% |       240 |
| 3.7      |   3.77% |       237 |
| 2.7      |   2.34% |       147 |
| 3.5      |   0.65% |        41 |
| 3.4      |   0.19% |        12 |
| 3.11     |   0.02% |         1 |
| Total    |         |     6,282 |

Source: `pip install -U pypistats && pypistats python_minor thefuck --last-month`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10"]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,14 +3,14 @@ name: Tests
 on: [push, pull_request]
 
 env:
-  PYTHON_LATEST: 3.9
+  PYTHON_LATEST: "3.10"
 
 jobs:
   test:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, 3.10-dev]
+        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10"]
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,13 +51,13 @@ flake8
 Run unit tests:
 
 ```bash
-py.test
+pytest
 ```
 
 Run unit and functional tests (requires docker):
 
 ```bash
-py.test --enable-functional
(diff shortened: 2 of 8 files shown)


import-brain (approved):

djh82:
Should this be rolled into a python 2 deprecation cycle - which I think has been mooted before?   I'd be much more in favour of working on removing python 2 support, especially since this change hasn't bought us much apart from no longer testing on python 3.5?

hugovk:
Dropping Python 2 is a good idea, it's been EOL for 2 years now.

I'd probably merge this first, because dropping 2.7 will be a bigger diff, so reviewing two PRs will be easier.

Python 3.6 is also now EOL, I'd recommend dropping that too.

hugovk:
In my experience, whilst many projects do bump major versions when dropping EOL versions (and I usually do too), modern pip and packaging means people can upgrade without things suddenly breaking.

This PR adds the necessary metadata for pip:

```python
      python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*',
```

This tells pip to only install this package for Python 2.7 and 3.6+.

People with Python 3.5 will stay on their current version.

I rarely see projects giving notice of dropping EOL Python versions. They're no longer receiving security updates (or any updates) from the core Python team. When you have the above metadata, upgrades go smoothly so no need to pre-announce.

But of course no problem if you prefer to give notice :)


---

And good timing :) The votes on this poll are currently pretty close over whether a major bump is needed. Interesting replies too.

image

https://twitter.com/nedbat/status/1476539511060447235

djh82:
Maybe when they drop python 2 support then? 😀

scorphus:
I'm in favor of it if we're talking about stating that we won't support Python 2 moving forward and removing it from the workflow and from `setup.py`.

But I think we should stop there. Maybe remove occurrences of `six`, or `try/import/except` or `sys.version*` or similar stuff. I don't see value in going over every single piece of the code and changing everything that was written to support 2 and 3 just for the sake of it.

hugovk:
Sounds good! Shall I include it here?

(By the way, https://github.com/asottile/pyupgrade/ is a really useful automated tool for upgrading syntax of an entire codebase.)

hugovk:
Closing this year old PR due to lack of interest, but happy to re-open if wanted.

PR #1253, opened 2021-12-14
Add instructions to use in corporate environment

### LINKED ISSUE #1252: Instructions for use in a corporate environment
Corporate environments often don't allow admin access to devices, and might have issues with swearing.

When I'm not at work, I'll do a proper PR to update the readme, but here's some notes for future-me:
```
pip3 install --user thefuck
```

```
⋊> ~ cat ~/Applications/thedang
#!/usr/bin/env python3
from thefuck.main import main
main()
```

```
thedang --alias | sed 's/fuck/dang/g' | sed 's/merge .*/merge/' > ~/.config/fish/functions/dang.fish
```

```
⋊> ~ ehco foo
fish: Unknown command: ehco
⋊> ~ dang
echo foo [enter/↑/↓/ctrl+c]
foo
```

mensly:
thefuck is extremely useful for developers and other professions that might require using, for example, a locked down Macbook that does not have local administrator access. I think it is important to include some instructions on how to get the tool set up in such an environment because it may not be obvious but fairly straightforward if you know what you're doing.

By the way, since I use it as an example, I got fish installed without admin access too:
```
ln -s ~/Applications/fish.app/Contents/Resources/base/usr/local/bin/fish ~/bin/fish
```

Fixes #1252

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,17 +543,29 @@ main()
 ```
 Lastly you need to configure your shell, follow the instructions to
 [configure for your choice of shell](https://github.com/nvbn/thefuck/wiki/Shell-aliases).
+If you have kept the name `thefuck`, this is simply the same as a standard install.
+Add the aliasing to your `.bash_profile`, `.bashrc`, `.zshrc` or other startup script
+```bash
+eval $(thefuck --alias)
+```
+
 If you have chosen an alternate name for the script you may need to use a
 tool like `sed` to modify the alias setup output to call the correct script. 
 As an example, here is how to configure fish to use run thefuck as `thedang` 
 with the `dang` alias:
 ```
 ⋊> ~ thedang --alias | sed 's/fuck/dang/g' > ~/.config/fish/functions/dang.fish
-⋊> ~ cat ~/.config/fish/functions/dang.fish | source
 ⋊> ~ ehco foo
 fish: Unknown command: ehco
 ⋊> ~ dang --yeah
 foo
+```
+
+Optionally, you may also want to edit the error message "No fucks given", as you have
+installed as a user the simplest way to do this would be to use to find the location 
+of the package and edit `ui.py`. eg:
+```bash
+sed -i 's/No fucks given/Oopsies!/' `pip3 show thefuck | grep 'Location: ' | sed 's/.* //'`/thefuck/ui.py
 ```
 
 ##### [Back to Contents](#contents)
--- a/README.md
+++ b/README.md
@@ -516,6 +516,48 @@ thefuck_contrib_foo
 
 ##### [Back to Contents](#contents)
 
+## Use in a Corporate Environment
+Installing thefuck is a lot easier when you have admin access to your machine,
+but many developers may be using computers owned by their employer.
+It is still possible to use thefuck without administrator access, but there 
+are some additional steps needed to install that only depend on you having a 
+python install.
+
+You might also like to take the opportunity to install it under an alternate
+name that is more acceptable to talk about out loud in the workplace such as
+Theph Oock, thefeck, or thedang.
+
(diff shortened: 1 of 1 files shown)


rany2:
I think a cleaner way than using `thedang --alias | sed ..` would be to just specify the alias to be dang:

```
user@ubuntu:~$ for x in ~/.local/bin/{fuck,thefuck}; do ln -vs "$x" "${x%fuck}dang"; done
'/home/user/.local/bin/dang' -> '/home/user/.local/bin/fuck'
'/home/user/.local/bin/thedang' -> '/home/user/.local/bin/thefuck'
user@ubuntu:~$ thedang --alias dang

            function dang () {
                TF_PYTHONIOENCODING=$PYTHONIOENCODING;
                export TF_SHELL=bash;
                export TF_ALIAS=dang;
                export TF_SHELL_ALIASES=$(alias);
                export TF_HISTORY=$(fc -ln -10);
                export PYTHONIOENCODING=utf-8;
                TF_CMD=$(
                    thefuck THEFUCK_ARGUMENT_PLACEHOLDER "$@"
                ) && eval "$TF_CMD";
                unset TF_HISTORY;
                export PYTHONIOENCODING=$TF_PYTHONIOENCODING;
                history -s $TF_CMD;
            }
        
user@ubuntu:~$ 
```

No need for sed. If you want you could use the above for the instructions for Bash.

mensly:
@rany2 The `--alias` argument only changes the name of the alias to type in, not the script that alias is calling. Take a look at `TF_CMD` in your example. This would've worked for you because `thefuck` is also in your `$PATH`, if someone has set this up as `thedang` instead of `thefuck` it would not work. `sed` will change both to remove the word "fuck".

rany2:
@mensly Sorry I didn't notice. I get why you did it now but I am not sure if `sed` might unintentionally break something in the future. Might be better to just add an option that allows you to set a custom `thefuck`

mensly:
This PR is intended only to document how the current product can be installed and used in a corporate environment, I like the idea of that option you spoke of though I think an alternative would be for the script name `thefuck` to be replaced with whatever the current command is that called `thefuck --alias`. Python isn't my main language, but I think `sys.argv[0]` has this string. I might experiment with this and if it proves useful do another PR after this one is merged.

I faced a lot of technical issues when I was trying to get this working myself and wanted to help others have an easier time. My current employer is actually ok with some casual swearing if it isn't used to hurt people, but I didn't know that when I was getting my computer set up which is why I set it up as dang. My example of configuring for fish is only an example of what is described in the previous paragraph. For a user without admin privileges that doesn't give a fuck about swearing, the configuration is a lot simpler and identical to that in the README already, ie adding `eval $(thefuck --alias)` to `.bashrc`. I'll add an extra sentence clarifying this.

As for outputting 'No fucks given', modifying the behaviour of the code will depend more on someone's environment. As this is something I haven't needed to do I can't tell you the best solution but I can imagine you could either edit your local copy of the `.py` file this string is contained in, or modifying the alias to edit stdout:
```
TF_CMD=$(
    thedang THEFUCK_ARGUMENT_PLACEHOLDER "$@" 2>&1 | sed 's/fuck/dang/'
) && eval "$TF_CMD";
```
A command line argument to set the error message (both when aliasing and evaluating) would be a better solution, so I might also look into that in the future PR I mentioned. To reiterate, I wanted to document what I needed to do to get the current version working on a corporate-controlled computer but am interested in modifying the software. I have another project I'm focused on at the moment, but worst case scenario this is MIT licensed so I can always publish to PyPI under the name `thedang` just to make it easier for people like me to use their preferred tools in their day job.

scorphus:
Wiki is the place for that.

PR #1258, opened 2022-01-03
fix: fish history file location

### LINKED ISSUE #1027: Fish shell does not appear to have its history rewritten
Fish shell successfully corrects the previous command, but the history is not re-written to include the correction.

I have a solution which I will post a PR for, so it can be discussed.

mainrs:
Fixes #1027

However, the entries still contain both, the original `fuck` command and the corrected one. Not sure if this is intended behavior or not.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ pypandoc
 pytest-benchmark
 pytest-docker-pexpect
 twine
+xdg
--- a/thefuck/shells/fish.py
+++ b/thefuck/shells/fish.py
@@ -8,6 +8,7 @@ from ..conf import settings
 from ..const import ARGUMENT_PLACEHOLDER
 from ..utils import DEVNULL, cache
 from .generic import Generic
+from xdg import xdg_data_home
 
 
 @cache('~/.config/fish/config.fish', '~/.config/fish/functions')
@@ -83,7 +84,7 @@ class Fish(Generic):
             return command_script
 
     def _get_history_file_name(self):
-        return os.path.expanduser('~/.config/fish/fish_history')
+        return xdg_data_home().joinpath("fish").joinpath("fish_history")
 
     def _get_history_line(self, command_script):
         return u'- cmd: {}\n   when: {}\n'.format(command_script, int(time()))


requirements.txt
```diff
@@ -9,3 +9,4 @@ pypandoc
 pytest-benchmark
 pytest-docker-pexpect
 twine
+xdg
```
scorphus, line 12: I don't think the addition of a new dependency is worth in this case.

How about sticking to what's been done so far in the project:

https://github.com/nvbn/thefuck/blob/841e3f9e13f52747c18319b55c77ece8095df306/thefuck/utils.py#L227-L228

In absence of `XDG_DATA_HOME`, Fish defaults to `~/.local/share`.

scorphus (changes requested):
Thanks for contributing! Please consider my observation.

requirements.txt
mainrs: Good point. I haven't written that much Python code so I kept sticking to a library. I've pushed a change.

thefuck/utils.py
```diff
@@ -196,6 +196,37 @@ def _for_app(fn, command):
...
+    # Ensure the cache_path exists, Python 2 does not have the exist_ok
+    # parameter
+    try:
+        os.makedirs(dir)
+    except OSError:
+        if not os.path.isdir(dir):
+            raise
+
+    return dir
+
```
djh82, line 221: Linting is probably going to expect 2 newlines between module functions.

mainrs: I couldn't check locally as I had trouble setting it up in nixos. would be nice if someone of you could fix this properly.

scorphus: Bull's eyes, @djh82.

@mainrs, you can check the GitHub Action output here: https://github.com/nvbn/thefuck/runs/4701882809?check_suite_focus=true (or clicking below in the "Some checks were not successful" section)

requirements.txt
scorphus: I understood that the `history` rule keeps suggesting `fuck (--yeah)`. If it's in shell's history and the `history` rule is suggesting it, then there might be an explanation for that. You can stick a breakpoint in the `match` function of `history` and inspect the input, it may help you understand that.

thefuck/utils.py
mainrs: Done!

tonytamps:
The requested changes appear to have been made. Would it be possible to re-review this MR?

Issue #1271, opened 2022-02-06
[Addition] Code of Conduct


FahimFBA:
The project is currently missing the Code of Conduct which defines standards for how to engage in a community. If you are interested in adding that, then I would want to add a PR with that. 

Please let me know what do you think about that.

jech:
Should the code of conduct forbid swearwords?

FahimFBA:
> Should the code of conduct forbid swearwords?
> 

Please let me add a Pull Request first. Then it would be helpful to discuss that,

RafaelKuhn:
well folks... Linus Torvalds was right

what the fuck man? you're PRing a code of conduct on the fucking thefuck project??
are you actually dysfunctionally retarded?

LoganDark:
> are you actually dysfunctionally retarded?

As described by this project's Code of Conduct this is not an acceptable way to address someone. Apologize immediately and also give me five dollars for taking the time to speak to you about this matter.

PR #1272, opened 2022-02-21
added the Code of Conduct

FahimFBA:
Closing #1271

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,133 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,133 @@
...
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
```
dannysauer, line 32: The intent of the phrase `sexualized language` clearly is not meant to include the name of the project, but for clarity's sake it might make sense to explicitly permit discussion which specifically refer to the project itself without any innuendo.  I don't have a great suggestion for specific verbiage, however.  And maybe people are smart enough to understand intent.  But in the linked issue, at least one commenter is clearly quite confused about language -- so I guess never underestimate how clueless people can be. :/

FahimFBA: > The intent of the phrase `sexualized language` clearly is not meant to include the name of the project, but for clarity's sake it might make sense to explicitly permit discussion which specifically refers to the project itself without any innuendo. I don't have a great suggestion for specific verbiage, however. And maybe people are smart enough to understand intent. But in the linked issue, at least one commenter is clearly quite confused about language -- so I guess never underestimate how clueless people can be. :/

Then should I remove the statement?

dannysauer: I don't personally have a strong opinion. Simply removing "language or" might be ok, or might not. I'd rather hear input from community leaders and not make that decision in a vacuum. :)

PR #1287, opened 2022-04-11
ls -r to ls -R

NikosKakonas:
The command `ls -R` is for cases where the user tries to unfuck `ls -r`. In such cases, the user intended to recursively list the current directory (that would happend if he had typed 'ls -R') but instead provided `-r`, which reverses the order of the list.


NikosKakonas:
I think that this rule does not need a test_match() function, as it unfucks ls -r, which does not has any output. This is because 'ls -r' is a valid command. So there are only 2 functions:

1st: test_not_match() and
2nd: test_get_new_command()

Am I correct?

I also add an entry on README.md for ls_r.

What do you think?

Issue #1282, opened 2022-04-01
It drops the quotes from a string when when it suggests a solution.


kxrn0:
If I fuck up with something like `gti commit -m "fist commit"`, it suggests `git commit -m first commit`, which is obviously wrong, since it drops the fucking quotes and fucks things up even more. Awesome fucking program, but fuck this shit, fr fr no cap bussin.

davidegavio:
I'd like to work on this issue, does anyone know where may I find the incriminated piece of code?

BrandonDoesDev:
@davidegavio If people knew that they'd fix it. Best to look for yourself or just add a feature that makes sure to include the quotes.

RVALLER:
Hey I'm trying to make certain rules to fix issues and I'm not sure what is meant by putting it in the ~/.config/thefuck/rules
Unless I'm just being silly, I don't know where to find ~/.config

scorphus:
Fixed in 5198b34f24ca4bc414a5bf1b0288ee86ea2529a8 (#1292). Thanks, @NikosKakonas and @djh82.

PR #1292, opened 2022-05-03
#1282 git misspelled

NikosKakonas:
The function git_support can not be useful if a command has the word "git" misspelled.
The purpose of this rule is to unfuck the commands like: gti commit -m 'message'.
Without this rule, for example, if you try to unfuck this command the result will be:
git commit -m message. This means that the quotation marks do not return. #1282

This rule is generally useful for every git command, in which the word git is misspelled.

I have also run some tests!


tests/rules/test_git_misspenlled.py
```diff
@@ -0,0 +1,26 @@
...
+
+@pytest.mark.parametrize(
+    "script, output, new_command",
+    [
+        ('''gti commit -m "this is a message"''', "", '''git commit -m "this is a message"'''),
+        ('''tgi commit -m "this is a message"''', "", '''git commit -m "this is a message"'''),
+    ],
+)
+def test_get_new_command(script, output, new_command):
+    assert get_new_command(Command(script, output)) == new_command
```
scorphus, line 26: Please, correct the name of this file.

Also, could you add `test_not_match` for cases where the rule does not match?

scorphus (approved):
This could be helpful for cases `no_command` isn't able to fix. Please, consider my suggestions below.

NikosKakonas:
I made test_not_match fucntion. I hope that now it is better!

djh82:
It feels like rather than solving this specifically for the git command, it might be prudent to solve the issue in the no_command rule?

NikosKakonas:
Thank you a lot for the review! Is this commit satisfying for the project?

djh82:
You'll probably want to revert the readme change and add some further tests for no_command?

scorphus:
Regarding the failure, please check #1310.

NikosKakonas:
@djh82 @scorphus
Are there any other mistakes of missing cases?

NikosKakonas:
I believe that now it is correct!

scorphus (changes requested):

scorphus:
@NikosKakonas, could you please rebase this on top of current `master`? Thanks!

NikosKakonas:
> @NikosKakonas, could you please rebase this on top of current `master`? Thanks!

image

@scorphus Could you please help me with this message?

NikosKakonas:
@scorphus  Is it now okay? Thanks!

NikosKakonas:
@scorphus Is it okay for merging?

scorphus (approved):
Thanks, @NikosKakonas! 👍

NikosKakonas:
The plesure is main!

alaycock:
@nkakonas does this justify a new release? This has been bugging me for a bit.

alaycock:
@scorphus could you cut a new release for this?

PR #1297, opened 2022-05-07
Upper to lowercase

NikosKakonas:
This rule convert the uppercase letters of the command to lowercase, only if all the letters of the command are uppercase (which is the most common mistake). 
For example, if I type
`$ LS`
an error will occur. So I will type
`$ fuck
ls ([enter/↑/↓/ctrl+c]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,7 @@ following rules are enabled by default:
 * `tmux` &ndash; fixes `tmux` commands;
 * `unknown_command` &ndash; fixes hadoop hdfs-style "unknown command", for example adds missing '-' to the command on `hdfs dfs ls`;
 * `unsudo` &ndash; removes `sudo` from previous command if a process refuses to run on superuser privilege.
+* `upper_to_lowercase` &ndash; replaces the uppercase letters with lowercase, if all the srcipt is typed in uppercase letters;
 * `vagrant_up` &ndash; starts up the vagrant instance;
 * `whois` &ndash; fixes `whois` command;
 * `workon_doesnt_exists` &ndash; fixes `virtualenvwrapper` env name os suggests to create new.
--- a/tests/rules/test_upper_to_lowercase.py
+++ b/tests/rules/test_upper_to_lowercase.py
@@ -1,0 +1,47 @@
+import pytest
+
+from thefuck.rules.upper_to_lowercase import match, get_new_command
+from thefuck.types import Command
+
+
+
+@pytest.mark.parametrize(
+    "script, output",
+    [
+        ("LS", ""),
+        ("CD THEFUCK", ""),
+        ("CAT README.MD", ""),
+        ("MV TESTS TESTING", ""),
+        ("GIT ADD .", ""),
+    ],
+)
+def test_match(script, output):
+    assert match(Command(script, output))
+
+
+@pytest.mark.parametrize(
+    "script, output",
+    [
+        ("ls", ""),
+        ("cd thefuck", ""),
+        ("cat README.md", ""),
+        ("mv tests testing", ""),
+        ("git add .", ""),
+    ],
+)
+def test_not_match(script, output):
+    assert not match(Command(script, output))
+
+
+@pytest.mark.parametrize(
+    "script, output, new_command",
+    [
+        ("LS", "", "ls"),
+        ("CD THEFUCK", "", "cd thefuck"),
+        ("CAT README.MD", "", "cat readme.md"),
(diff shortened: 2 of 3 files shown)


Rom1deTroyes:
This one is a nice idea, but what about `LS _A` ?

And, a watching point : env vars are upper case, and comes before the command :
```sh
FLASK_APP=app python -m flask run
```
(I don't know, if either of this points are relevant, just thinking :thinking: )

NikosKakonas:
Yes you are right, I will try to fix those too. I'm open to suggestions. Thanks!

NikosKakonas:
@scorphus I converted this pull request to a draft, so I can improve it as @Rom1deTroyes mentioned. I have already write some code. Do you think is a good idea?

thefuck/rules/upper_to_lower_case.py
```diff
@@ -0,0 +1,54 @@
...
+    all_correct_commands = []
+    for com in new_commands:
+        result = subprocess.getstatusoutput(com)
+        if result[0] == 0:
+            all_correct_commands.append(com)
+    return all_correct_commands
+
+
+def match(command):
+    flag = False
```
djh82, line 35: A name like 'matches' might be more typical; flag is not very descriptive

thefuck/rules/upper_to_lower_case.py
```diff
@@ -0,0 +1,54 @@
...
+    for i in command.script:
+        if i.isupper():
+            flag = True
```
djh82, line 38: Isn't this only set if the last character is uppercase?

thefuck/rules/upper_to_lower_case.py
```diff
@@ -0,0 +1,54 @@
...
+        new_command = new_command.strip()
+        if new_command not in new_commands:
+            new_commands.append(new_command)
+    return new_commands
+
+
+def find_all_correct_commands(new_commands : list):
+    all_correct_commands = []
+    for com in new_commands:
+        result = subprocess.getstatusoutput(com)
```
djh82, line 28: I don't think we want to actually run the command, this seems really dangerous!  What if the command is autocorrected to be an 'rm' for instance?

thefuck/rules/upper_to_lower_case.py
```diff
@@ -0,0 +1,54 @@
...
+
+def match(command):
+    flag = False
+    for i in command.script:
+        if i.isupper():
+            flag = True
+    # checks if the command that typed is valid, if it is not it retunrs false
+    # So if we type: SL -A (instead of LS -A) the rule will not be activated
+    new_commands = find_all_possible_commands(command)
+    all_correct_commands = find_all_correct_commands(new_commands)
```
djh82, line 42: Are we trying to be to clever here?  Could we just offer the tolowercase rules?

djh82:
I have a much simpler implementation as part of my local rules, which seems to catch most of my issues, and doesnt try to do too much:

```python
from thefuck.utils import get_all_executables, which


def match(command):
    return (command.script.isupper()
    	    and not which(command.script_parts[0])
            and ('not found' in command.output
                 or 'is not recognized as' in command.output)
            and (command.script_parts[0].lower() in get_all_executables()))


def get_new_command(cmd):
    return cmd.script.lower()


requires_output = False
priority = 100
```

thefuck/rules/upper_to_lower_case.py
NikosKakonas: Do you know how am I going to check if a command is valid or not? (without running it)

thefuck/rules/upper_to_lower_case.py
NikosKakonas: You are right I didn't know that there was already the get_all_executables function in thefuck.utils

thefuck/rules/upper_to_lower_case.py
NikosKakonas: Okay

thefuck/rules/upper_to_lower_case.py
NikosKakonas: Should I set the variable "matches" = True only if all the characters of the command are upper case?

thefuck/rules/upper_to_lower_case.py
NikosKakonas: Yes this is very dangerous! I agree.

NikosKakonas:
Thank you very much for reviewing! I followed your instructions.

PR #1298, opened 2022-05-16
Prioritise main/master mix ups

arjentz:
Fixing the mix up of branches `main` and `master` should have a higher priority than other rules, because it is more logical most of the time. One example:
```
$ git checkout main
error: pathspec 'main' did not match any file(s) known to git
```

The suggestion `git checkout master` should have the highest priority, rather than `git checkout -b main` which is the case now.

### DIFF
--- a/thefuck/rules/git_main_master.py
+++ b/thefuck/rules/git_main_master.py
@@ -13,4 +13,4 @@ def get_new_command(command):
     return command.script.replace("main", "master")
 
 
-priority = 1200
+priority = 800


djh82:
Presumably this can be closed now too?

scorphus (commented):
Thanks for contributing, @arjentz.

This rule matches ***any*** command that has either `"main"` or `"master"` in its output. For that reason, it shouldn't be matched before ***all*** other rules.

According to [the docs](https://github.com/nvbn/thefuck/#settings), for a special case where a user wants a specific rule to match before others, a special entry in `~/.config/thefuck/settings.py` suffices:

```python
# ...
priority = {
    "git_main_master": 999,
    # ...
}
# ...
```

Having wrote that, this PR won't be accepted.

PR #1302, opened 2022-05-24
New rule: fix missing `git clone` when given a URL

MiguelGuthridge:
Often when I'm trying to `git clone` a project, I'll paste the URL into my terminal expecting it to also contain the `git clone` part. This rule should allow `thefuck` to detect the missing `git clone` and suggest it when someone pastes an SSH or HTTP/HTTP url that ends in `.git`.

A rule for correcting for double `git clone git clone [repo]` already exists. This rule addresses the opposite problem.

image

I've also added some simple tests for it, which appear to pass.

This is my first PR with any meaningful change, so I'd appreciate some feedback on it. It still needs a bit of cleaning up, but I think it should be good to merge in once I remove a couple of random things

### DIFF
--- a/tests/rules/test_git_clone_missing.py
+++ b/tests/rules/test_git_clone_missing.py
@@ -22,20 +22,10 @@ invalid_urls = [
     'https:/github.com/nvbn/thefuck.git'  # Bad protocol
 ]
 
-# TODO: Powershell,
-shell_errors = [
-    'not found',  # sh
-    'command not found',  # fish
-    'No such file or directory',  # bash
-    'no such file or directory',  # zsh
-    'is not recognized as',  # cmd, powershell
-]
-
-
 
 @pytest.mark.parametrize(
     'cmd',
-    [Command(c, e) for c, e in product(valid_urls, shell_errors)]
+    [Command(c, 'not found') for c in valid_urls]
 )
 def test_match(cmd):
     assert match(cmd)
@@ -43,7 +33,7 @@ def test_match(cmd):
 
 @pytest.mark.parametrize(
     'cmd',
-    [Command(c, shell_errors[0]) for c in invalid_urls]
+    [Command(c, 'not found') for c in invalid_urls]
 )
 def test_not_match(cmd):
     assert not match(cmd)
@@ -51,7 +41,7 @@ def test_not_match(cmd):
 
 @pytest.mark.parametrize(
     'cmd',
-    [Command(c, shell_errors[0]) for c in valid_urls]
+    [Command(c, 'not found') for c in valid_urls]
 )
 def test_get_new_command(cmd):
     assert get_new_command(cmd) == 'git clone ' + cmd.script
--- a/tests/rules/test_git_clone_missing.py
+++ b/tests/rules/test_git_clone_missing.py
@@ -1,0 +1,37 @@
+from thefuck.rules.git_clone_git_clone import match, get_new_command
+from thefuck.types import Command
+
+output_clean = """
+fatal: Too many arguments.
+
+usage: git clone [<options>] [--] <repo> [<dir>]
+"""
+
+http_url = 'https://github.com/nvbn/thefuck.git'
+ssh_url = 'git@github.com:nvbn/thefuck.git'
+
+# HELP WANTED: How can I make this independent of the shell?
+http_output = """
+bash: https://github.com/nvbn/thefuck.git: No such file or directory
+"""
+ssh_output = """
+bash: hgit@github.com:nvbn/thefuck.git: No such file or directory
+"""
+
+
+def test_match():
+    assert match(Command(http_url, http_output))
+    assert match(Command(ssh_url, ssh_output))
+
+
+def test_not_match():
+    assert not match(Command('', ''))
+    assert not match(Command('git branch', ''))
+    assert not match(Command('git clone foo', ''))
+    assert not match(Command('git clone foo bar baz', output_clean))
+    assert not match(Command('git clone ' + http_url, ''))
+
+
+def test_get_new_command():
+    assert get_new_command(Command(http_url, http_output)) == 'git clone ' + http_url
+    assert get_new_command(Command(ssh_url, ssh_output)) == 'git clone ' + ssh_url
--- a/tests/rules/test_git_clone_missing.py
+++ b/tests/rules/test_git_clone_missing.py
@@ -13,11 +13,12 @@ valid_urls = [
 invalid_urls = [
     '',  # No command
     'notacommand',  # Command not found
+    'ssh git@github.com:nvbn/thefrick.git',  # ssh command, not a git clone
     'git clone foo',  # Valid clone
     'git clone https://github.com/nvbn/thefuck.git',  # Full command
     'github.com/nvbn/thefuck.git',  # Missing protocol
     'github.com:nvbn/thefuck.git',  # SSH missing username
-    'git clone git clone ssh://git@github.com:nvbn/thefrick.git',  # Double clone
+    'git clone git clone ssh://git@github.com:nvbn/thefrick.git',  # 2x clone
     'https:/github.com/nvbn/thefuck.git'  # Bad protocol
 ]
 
@@ -38,9 +39,8 @@ def test_not_match(cmd):
     assert not match(cmd)
 
 
-@pytest.mark.parametrize(
-    'cmd',
-    [Command(c, 'not found') for c in valid_urls]
-)
-def test_get_new_command(cmd):
-    assert get_new_command(cmd) == 'git clone ' + cmd.script
+@pytest.mark.parametrize('script', valid_urls)
+def test_get_new_command(script):
+    command = Command(script, 'not found')
+    new_command = 'git clone ' + script
+    assert get_new_command(command) == new_command
--- a/tests/rules/test_git_clone_missing.py
+++ b/tests/rules/test_git_clone_missing.py
@@ -1,53 +1,60 @@
 import pytest
+from itertools import product
 from thefuck.rules.git_clone_missing import match, get_new_command
 from thefuck.types import Command
 
-https_url = 'https://github.com/nvbn/thefuck.git'
-http_url = 'http://github.com/nvbn/thefuck.git'
-ssh_url = 'git@github.com:nvbn/thefuck.git'
+valid_urls = [
+    'https://github.com/nvbn/thefuck.git',
+    'https://github.com/nvbn/thefuck',
+    'http://github.com/nvbn/thefuck.git',
+    'git@github.com:nvbn/thefuck.git',
+    'git@github.com:nvbn/thefuck',
+    'ssh://git@github.com:nvbn/thefuck.git',
+]
+invalid_urls = [
+    '',  # No command
+    'notacommand',  # Command not found
+    'git clone foo',  # Valid clone
+    'git clone https://github.com/nvbn/thefuck.git',  # Full command
+    'github.com/nvbn/thefuck.git',  # Missing protocol
+    'github.com:nvbn/thefuck.git',  # SSH missing username
+    'git clone git clone ssh://git@github.com:nvbn/thefrick.git',  # Double clone
+    'https:/github.com/nvbn/thefuck.git'  # Bad protocol
+]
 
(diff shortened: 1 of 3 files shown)


thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,26 @@
+from thefuck.specific.git import git_support
+
+output_contains = ': not found'
```
djh82, line 3: Fish returns `repo.git : command not found`

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,26 @@
...
+# Note: could probably use better checking for URLs
+
+
+# @git_support
```
djh82, line 7: this commented out line can be removed

tests/rules/test_git_clone_missing.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+from thefuck.rules.git_clone_missing import match, get_new_command
+from thefuck.types import Command
+
+https_url = 'https://github.com/nvbn/thefuck.git'
+http_url = 'http://github.com/nvbn/thefuck.git'
+ssh_url = 'git@github.com:nvbn/thefuck.git'
```
djh82, line 7: SSH cloning might also look like `ssh://git@github`, at which point I dont think your match rule will work as expected

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+        or (  # SSH cloning: user[@]website.com[:]path/to/repo[.git]
+            command.script.find('@') < command.script.find(':')
+            and command.script.find('@') != -1
+            and command.script.find(':') != -1
+        )
+    )
+
+
+# @git_support
```
djh82, line 24: remove me

thefuck/rules/git_clone_missing.py
djh82: Base returns: `: No such file or directory`

djh82: So you might consider searching for `not found` or `no such file or directory`?

MiguelGuthridge: Fixed with refactor

thefuck/rules/git_clone_missing.py
MiguelGuthridge: Fixed :)

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,26 @@
+from thefuck.specific.git import git_support
+
+output_contains = ': not found'
+# Note: could probably use better checking for URLs
+
+
+# @git_support
+def match(command):
+    if not output_contains in command.output:
```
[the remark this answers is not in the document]
MiguelGuthridge: Fixed with refactor

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,26 @@
...
+        return False
+    return command.script.endswith('.git') and (
```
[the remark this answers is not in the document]
MiguelGuthridge: Good idea, implemented this. Turns out `urllib` is a bit fussy with SSH things so I had to do a bunch of yuckiness to get it working but it seems to work now

tests/rules/test_git_clone_missing.py
MiguelGuthridge: Fixed

MiguelGuthridge:
@djh82 would appreciate a re-review. I think I've fixed up all the issues

scorphus (commented):
This rule should be output-independent. Why bother matching the output?

However, it reads quite specific to me, I don't think many users would benefit from it. A URL might be related to so many things, not only `git clone`.

The Fuck allows the user to have user-defined rules at their disposal: [Creating your own rules](https://github.com/nvbn/thefuck/#creating-your-own-rules).

Not every single rule has to be part of The Fuck.

MiguelGuthridge:
@scorphus thanks for your feedback! I believe that this rule would be worthwhile as I encountered it enough times to choose to spend my time on creating a rule for it. I tutor a computer science class where we teach git and have observed that I'm far from the only person to make this mistake. Since there's a rule for handling `git clone git clone`, which is useful when I type `git clone` then paste the URL, only for it to include its own `git clone` (Bitbucket does this), I'd say that this particular rule is reasonable enough, since it handles the opposite case.

Perhaps it could search for `'git'` within the URL as an additional measure to avoid false positives? This would still work for all most of the popular git servers I'm aware of, at least using SSH. Would that change make the addition be worthwhile? Can you think of any other methods I could use to reduce the number of false positives?

scorphus:
Thanks for sharing your testimonial. Let's get this merged.

Regarding matching the output, I thought about disregarding output altogether and caring only about `output.script` or `output.script_parts`.

Asserting `git` in the URL would blacklist bitbucket repos.

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,42 @@
...
+
+Correct missing `git clone` command when pasting a git URL
+
+```sh
+& https://github.com/nvbn/thefuck.git
+git clone https://github.com/nvbn/thefuck.git
+
+Author: Miguel Guthridge
+'''
+from urllib import parse
```
scorphus, line 12: We still support py2:

```suggestion
from six.moves.urllib import parse
```

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,42 @@
...
+    'is not recognized as',
+]
+
+
+def match(command):
+    # Ignore capitalisation in output
+    output = command.output.lower()
+    script = command.script
+    # Check for a command not found error
+    if not any([search in output for search in output_contains]):
```
scorphus, line 26: Let's not create a list just to iterate over it:

```suggestion
    if not any(search in output for search in output_contains):
```

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,42 @@
...
+        return False
+    # URLs can't have spaces
+    if ' ' in command.script:
```
scorphus, line 29: Let's avoid iterating over that string once again:
```suggestion
    if len(command.script_parts) > 1:
```

tests/rules/test_git_clone_missing.py
```diff
@@ -0,0 +1,60 @@
...
+]
+
+
+
+@pytest.mark.parametrize(
+    'cmd',
+    [Command(c, e) for c, e in product(valid_urls, shell_errors)]
+)
+def test_match(cmd):
+    print(cmd)
```
scorphus, line 41:
```suggestion
```

tests/rules/test_git_clone_missing.py
```diff
@@ -0,0 +1,60 @@
...
+    assert match(cmd)
+
+
+@pytest.mark.parametrize(
+    'cmd',
+    [Command(c, shell_errors[0]) for c in invalid_urls]
+)
+def test_not_match(cmd):
+    print(cmd)
```
scorphus, line 50:
```suggestion
```

tests/rules/test_git_clone_missing.py
```diff
@@ -0,0 +1,60 @@
...
+    assert not match(cmd)
+
+
+@pytest.mark.parametrize(
+    'cmd',
+    [Command(c, shell_errors[0]) for c in valid_urls]
+)
+def test_get_new_command(cmd):
+    print(cmd)
```
scorphus, line 59:
```suggestion
```

scorphus (changes requested):
Thanks for contributing, please consider my suggestions below.

Also, a new line in README is required.

MiguelGuthridge:
Thanks again for the feedback! I fixed up things as per your suggestion, and also removed the code that checked the output.

Double checking, since I'm not entirely sure, should it check the URL for the presence of `git` somewhere? The only case I can think of that wouldn't work would be cloning using HTTPS on Bitbucket where `.git` has been removed from the end of their copied URL. Their SSH username is `git`, and their copy clone button copies the `git clone` part anyway. It would be helpful for preventing false positives such as [YouTube videos](https://www.youtube.com/watch?v=dQw4w9WgXcQ).

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,33 @@
...
+```
+
+Author: Miguel Guthridge
+'''
+from six.moves.urllib import parse
+
+
+def match(command):
+    script = command.script
+    parts = command.script_parts
```
scorphus, line 18: No need to care about Bash nor ZSH nor Fish, only `sh` and `cmd`/`pwsh`:
```suggestion
    # Check for a command not found error
    if which(command.script_parts[0]) or not (
        'not found' in command.output or 'is not recognized as' in command.output
    ):
        return False
    script = command.script
```
It's also marginally faster than using a generator with `any`.

Nit: declare variables only when necessary.

thefuck/rules/git_clone_missing.py
```diff
@@ -0,0 +1,33 @@
...
+    # We want it to be a URL by itself
+    if len(parts) > 1:
```
scorphus, line 20: No need to create a new variable, that's readable enough:
```suggestion
    if len(command.script_parts) > 1:
```

tests/rules/test_git_clone_missing.py
```diff
@@ -0,0 +1,47 @@
...
+def test_not_match(cmd):
+    assert not match(cmd)
+
+
+@pytest.mark.parametrize(
+    'cmd',
+    [Command(c, 'not found') for c in valid_urls]
+)
+def test_get_new_command(cmd):
+    assert get_new_command(cmd) == 'git clone ' + cmd.script
```
scorphus, line 47: Would be better if the `Command` instance was created in the test itself rather than in the parameterization. Then `script` could be used instead of `cmd.script`:

```suggestion
@pytest.mark.parametrize('script', valid_urls)
def test_get_new_command(script):
    command = Command(script, shell_errors[0])
    new_command = 'git clone ' + script
    assert get_new_command(command) == new_command
```

tests/rules/test_git_clone_missing.py
```diff
@@ -0,0 +1,47 @@
...
+def test_match(cmd):
+    assert match(cmd)
+
+
+@pytest.mark.parametrize(
+    'cmd',
+    [Command(c, 'not found') for c in invalid_urls]
+)
+def test_not_match(cmd):
+    assert not match(cmd)
```
scorphus, line 39: Ditto.

scorphus (changes requested):
My previous review was partially written. I would conclude the thought that I initiated in the message before, but ended up forgetting about it. Apologies.

I'd complement by saying that checking the command output is actually desirable and I should've given it a bit more thought when commenting earlier. Doing so speeds things up for cases that don't match by a factor of 4, at least.

But I think it should be simplified. Please consider my suggestions below.

Also, please, could you bring back the tests cases that include the output? Thanks!

MiguelGuthridge:
Pretty sure I've got everything fixed up now! Let me know if there's anything else you'd like me to do!

MiguelGuthridge:
Just bumping this one again @scorphus 
Is there anything else you want me to do on it?

scorphus (approved):
Sorry for the delay, @MiguelGuthridge.

> Pretty sure I've got everything fixed up now! Let me know if there's anything else you'd like me to do!

You nailed it. :+1:

My bad: I've just realized that, for commands like `git@github.com:nvbn/thefuck.git` and `https://github.com/nvbn/thefuck.git`, [Macos's `sh`](# "GNU bash, version 3.2.57(1)-release-(x86_64-apple-darwin19)") errs with `/bin/sh: https://github.com/nvbn/thefuck.git: No such file or directory\n`.

I'll got ahead and make the necessary change and push it on top of yours. Let's get this merged.

scorphus:
@MiguelGuthridge, please let me know your thoughts about the last change I submitted.

MiguelGuthridge:
Looks pretty great to me!

scorphus (approved):
Thank you, @MiguelGuthridge! :shipit:

scorphus:
@MiguelGuthridge, thanks for the complete patch! Much appreciated! 🙌

PR #1308, opened 2022-05-29
Added updatedb (mlocate) to sudo rules

0xphk:
Added the pattern for ```updatedb``` (mlocate), if command is run without ```sudo``` privileges, to ```rules/sudo.py```


scorphus (approved):
Thanks, @0xphk

PR #1317, opened 2022-07-01
Add terraform 'no command' rule

josepdaniel:
Every single time I use terraform to destroy resources I type `terraform destory` instead 🙃. This adds a rule to fix those simple mistakes.

### DIFF
--- a/tests/rules/test_terraform_no_command.py
+++ b/tests/rules/test_terraform_no_command.py
@@ -4,8 +4,8 @@ from thefuck.types import Command
 
 
 @pytest.mark.parametrize('script, output', [
-    ('terraform appyl', 'Terraform has no command named \"appyl\". Did you mean \"apply\"?)'),
-    ('terraform destory', 'Terraform has no command named \"destory\". Did you mean \"destroy\"?)')])
+    ('terraform appyl', 'Terraform has no command named "appyl". Did you mean "apply"?'),
+    ('terraform destory', 'Terraform has no command named "destory". Did you mean "destroy"?')])
 def test_match(script, output):
     assert match(Command(script, output))
 
@@ -16,12 +16,12 @@ def test_match(script, output):
     ('terraform apply', 'Apply complete! Resources: 0 added, 0 changed, 0 destroyed.'),
 ])
 def test_not_match(script, output):
-    assert not match(Command(script, output=output))
+    assert not match(Command(script, output))
 
 
-@pytest.mark.parametrize('command, new_command', [
-    (Command('terraform appyl', 'Terraform has no command named \"appyl\". Did you mean \"apply\"?)'), 'terraform apply'),
-    (Command('terraform destory --some-other-option', 'Terraform has no command named \"destory\". Did you mean \"destroy\"?)'), 'terraform destroy --some-other-option'),
+@pytest.mark.parametrize('script, output, new_command', [
+    ('terraform appyl', 'Terraform has no command named "appyl". Did you mean "apply"?', 'terraform apply',),
+    ('terraform destory --some-other-option', 'Terraform has no command named "destory". Did you mean "destroy"?', 'terraform destroy --some-other-option',),
 ])
-def test_get_new_command(command, new_command):
-    assert get_new_command(command) == new_command
+def test_get_new_command(script, output, new_command):
+    assert get_new_command(Command(script, output)) == new_command
--- a/tests/rules/test_terraform_no_command.py
+++ b/tests/rules/test_terraform_no_command.py
@@ -1,0 +1,27 @@
+import pytest
+from thefuck.rules.terraform_no_command import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('script, output', [
+    ('terraform appyl', 'Terraform has no command named \"appyl\". Did you mean \"apply\"?)'),
+    ('terraform destory', 'Terraform has no command named \"destory\". Did you mean \"destroy\"?)')])
+def test_match(script, output):
+    assert match(Command(script, output))
+
+
+@pytest.mark.parametrize('script, output', [
+    ('terraform --version', 'Terraform v0.12.2'),
+    ('terraform plan', 'No changes. Infrastructure is up-to-date.'),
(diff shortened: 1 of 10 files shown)


scorphus:
Wonder why the workflow didn't run. I changed it to include reopened PRs, and there we have it.

tests/rules/test_terraform_no_command.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+from thefuck.rules.terraform_no_command import match, get_new_command
+from thefuck.types import Command
+
+
+@pytest.mark.parametrize('script, output', [
+    ('terraform appyl', 'Terraform has no command named \"appyl\". Did you mean \"apply\"?)'),
+    ('terraform destory', 'Terraform has no command named \"destory\". Did you mean \"destroy\"?)')])
```
scorphus, line 8:
```suggestion
    ('terraform appyl', 'Terraform has no command named "appyl". Did you mean "apply"?'),
    ('terraform destory', 'Terraform has no command named "destory". Did you mean "destroy"?')])
```

tests/rules/test_terraform_no_command.py
```diff
@@ -0,0 +1,27 @@
...
+    assert match(Command(script, output))
+
+
+@pytest.mark.parametrize('script, output', [
+    ('terraform --version', 'Terraform v0.12.2'),
+    ('terraform plan', 'No changes. Infrastructure is up-to-date.'),
+    ('terraform apply', 'Apply complete! Resources: 0 added, 0 changed, 0 destroyed.'),
+])
+def test_not_match(script, output):
+    assert not match(Command(script, output=output))
```
scorphus, line 19:
```suggestion
    assert not match(Command(script, output))
```

tests/rules/test_terraform_no_command.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.mark.parametrize('command, new_command', [
+    (Command('terraform appyl', 'Terraform has no command named \"appyl\". Did you mean \"apply\"?)'), 'terraform apply'),
+    (Command('terraform destory --some-other-option', 'Terraform has no command named \"destory\". Did you mean \"destroy\"?)'), 'terraform destroy --some-other-option'),
+])
+def test_get_new_command(command, new_command):
+    assert get_new_command(command) == new_command
```
scorphus, line 27: I like that you passed in string parameters and instantiated `Command` with them. How about doing the same here?

```suggestion
@pytest.mark.parametrize('script, output, new_command', [
    ('terraform appyl', 'Terraform has no command named "appyl". Did you mean "apply"?', 'terraform apply',),
    ('terraform destory --some-other-option', 'Terraform has no command named "destory". Did you mean "destroy"?', 'terraform destroy --some-other-option',),
])
def test_get_new_command(script, output, new_command):
    assert get_new_command(Command(script, output)) == new_command
```

thefuck/rules/terraform_no_command.py
```diff
@@ -0,0 +1,16 @@
+import re
+from thefuck.utils import for_app
+
+MISTAKE = r"(?<=Terraform has no command named \")(.*)(?=\"\.)"
+FIX = r"(?<=Did you mean \")(.*)(?=\"\?)"
```
scorphus, line 5: How about being a bit more restrictive and remove some unnecessary backslashes?

```suggestion
MISTAKE = r'(?<=Terraform has no command named ")([^"]+)(?="\.)'
FIX = r'(?<=Did you mean ")([^"]+)(?="\?)'
```

thefuck/rules/terraform_no_command.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@for_app('terraform')
+def match(command):
+    return ('Terraform has no command named' in command.output and 'Did you mean' in command.output)
+
+
+def get_new_command(command):
+    mistake = re.search(MISTAKE, command.output).group(0)
+    fix = re.search(FIX, command.output).group(0)
```
scorphus, line 15: These lines can fail if the structure of the output message changes a bit — such as punctuation. If that ever happens, the rule will start failing. But that's better than it silently stop matching because of a stricter `match` function.

That hints at the possibility of running each rule inside a try/except block and displaying a proper error instead of letting it explode. But I digress (note taken).

scorphus (commented):
Thanks for the new rule, @josepdaniel! It looks great. Also thanks for the complete patch, greatly appreciated.

Please, consider a few suggestions below.

josepdaniel:
hey @scorphus, thanks for the comments. Made the requested changes, let me know if you have any more thoughts.

thefuck/rules/terraform_no_command.py
josepdaniel: Ok, I changed the `match` method to only return positive if both regex patterns can be matched against the command output, so `get_new_command` hypothetically won't ever fail.

scorphus: I'm not sure whether I'd prefer having it fail or not. This way it wouldn't silently fail to match if the output ever changes in the future.

The failure would have to be properly handled (try/except) — which can be done separately. With that, I think it should fail. The failure will be logged and thefuck would jump to the next rule.

Without it being handled, which is the current behavior, thefuck would crash and any following rules would not be attempted ( namely their `get_new_command`).

To me, a failing rule is way more noticeable that a silent, non-matching one.

WDYT?

scorphus:
Awesome! Would you please merge with current `master` again? TIA

thefuck/rules/terraform_no_command.py
josepdaniel: I think tracking a moving target can be tough - just how lenient should the `match` function be in order for the rule to continue matching when the output changes? It could be problematic to ask the writers of the rules to come up with a match function that is broad enough to capture possible evolutions of the error message while remaining specific to the problem it is addressing. 

My thought process in the moment was that 'if the rule matches then `fuck` should be able to provide the replacement command'. Either way, I would be happy to revert back to the original matching logic if you feel strongly about it

scorphus (approved):
Hey, @josepdaniel, apologies for the delay. Thanks for contributing! :shipit:

josepdaniel:
Cheers!

PR #1344, opened 2022-11-21
Use unittest's buildin mock

jelly:
Since Python 3.3 unittest has a mock module buildin, so a seperate mock module is no longer required.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 flake8
 pytest
-mock
 pytest-mock
 wheel
 setuptools>=17.1
--- a/tests/entrypoints/test_alias.py
+++ b/tests/entrypoints/test_alias.py
@@ -1,4 +1,4 @@
-from mock import Mock
+from unittest.mock import Mock
 import pytest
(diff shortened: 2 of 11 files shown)


deepyaman:
> Since Python 3.3 unittest has a mock module buildin, so a seperate mock module is no longer required.

Since `pytest-mock` is already being used, why not just use `mocker` directly? For example, the below line can just use `mocker.Mock` and ditch the import altogether:

https://github.com/nvbn/thefuck/blob/b3da68abb96300ca7f0a42d1b6f430fccd6349f1/tests/entrypoints/test_alias.py#L18

jelly:
> > Since Python 3.3 unittest has a mock module buildin, so a seperate mock module is no longer required.
> 
> Since `pytest-mock` is already being used, why not just use `mocker` directly? For example, the below line can just use `mocker.Mock` and ditch the import altogether:
> 
> https://github.com/nvbn/thefuck/blob/b3da68abb96300ca7f0a42d1b6f430fccd6349f1/tests/entrypoints/test_alias.py#L18

If the maintainer wants this rather sure, my intention was to just drop the `mock` dependency.

PR #1348, opened 2022-12-09
fix: add tests folder to .gitignore

jl-houss:
i've seen that there's a tests folder which has been added to the project and i guess it was by mistake

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ __pycache__/
 .Python
 env/
 build/
+tests/
 develop-eggs/
 dist/
 downloads/


Rom1deTroyes:
The `/tests/` directory is THE place where application is tested :1st_place_medal: 

For more information on testing a Python app : https://realpython.com/python-testing/

Welcome in the world of TDD :-)

jl-houss:
so it's meant to be in the master branch release?

deepyaman:
@jl-houss What is or isn't in the released package is defined in `setup.py`. You can see that `tests` are excluded from there: https://github.com/nvbn/thefuck/blob/master/setup.py#L51-L52

PR #1363, opened 2023-03-02
Add ChatGPT as a rule, disabled by default.

XieGuochao:
This patch enhances thefuck by querying ChatGPT optionally. The user sets an environment variable `THEFUCK_OPENAI_TOKEN` from OpenAI to enable it.

Three new arguments are added:
1. `-c` or `--chatgpt`: the number of suggestions from ChatGPT. If it is 0, ChatGPT is disabled.
2. `-t` or `--chatgpt-token`: the maximum tokens for a ChatGPT query. By default, it is 100.
3. `-m` or `--chatgpt-model`: the ChatGPT model used. By default, it is `gpt-3.5-turbo`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,10 @@ default:
 * `git_push_force` &ndash; adds `--force-with-lease` to a `git push` (may conflict with `git_push_pull`);
 * `rm_root` &ndash; adds `--no-preserve-root` to `rm -rf /` command.
 
+The following rule uses OpenAI ChatGPT. To enable it, you need to set the environment variable `THEFUCK_OPENAI_TOKEN=<OpenAI Token>` and pass in `--chatgpt [No. SUGGESTIONS >= 1]`:
+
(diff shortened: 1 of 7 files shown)


pannous:
Fantastic PR 

police merch (please merge!)

why not use OPENAI_TOKEN directly / as fallback?

pannous:
100 token default limits seems pretty low though?

pannous:
--chatgpt-token should be --chatgpt-max-tokens for clarity

--chatgpt-model can probably be removed since a new model will likely result in a new API

XieGuochao:
> Fantastic PR
> 
> police merch (please merge!)
> 
> why not use OPENAI_TOKEN directly / as fallback?

I was worrying users accidentally enabled it while one was using other OpenAI services. I am totally fine using `OPENAI_TOKEN` directly / as fallback.



> 100 token default limits seems pretty low though?

Great point! I am not a ChatGPT expert and can you suggest a better limit?



> --chatgpt-token should be --chatgpt-max-tokens for clarity

Agree!

> --chatgpt-model can probably be removed since a new model will likely result in a new API

I don't know if there would be GPT-4.0 or something similar with the same API but more data...

xingjian-zhang:
This looks so promising to me. Like the idea. 🚀

thibaultmol:
Status of this PR?

PR #1366, opened 2023-03-07
Add rule to use terraform help text for suggestions

rpdelaney:
Using the approach in #1361 to solve the same problem with terraform. 

https://regex101.com/r/kJ4rEC/1

### DIFF
--- a/thefuck/rules/terraform_help.py
+++ b/thefuck/rules/terraform_help.py
@@ -1,0 +1,34 @@
+#
+# $ terraform destory
+#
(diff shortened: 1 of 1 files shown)


rpdelaney:
Related: #1317 
cc @josepdaniel

rpdelaney:
Ah, I misunderstood what #1317 does -- this does pretty much the same thing. I think this can probably be closed, but I'll leave this here in case the small differences in my approach are preferred.

PR #1371, opened 2023-04-05
Added basic support for Atuin

StandingPadAnimations:
Atuin is shell history on steriods and is commonly used as an enhancement of the existing shell history. Unlike Bash or Zsh, Atuin uses an SQLite database to store the command history.

The nice thing about this however is that this makes Atuin shell agnostic, so all we've done is add a basic check in the generic shell class to check to see if the user provided a path to the Atuin database. If so, then we check the database instead of the default location.

Right now this system has an issue where it'll reconnect to the database multiple times, which may cause performance issues. The point however is that this works, and that we can search the Atuin database.

### DIFF
--- a/thefuck/const.py
+++ b/thefuck/const.py
@@ -44,7 +44,8 @@ DEFAULT_SETTINGS = {'rules': DEFAULT_RULES,
                     'instant_mode': False,
                     'num_close_matches': 3,
                     'env': {'LC_ALL': 'C', 'LANG': 'C', 'GIT_TRACE': '1'},
-                    'excluded_search_path_prefixes': []}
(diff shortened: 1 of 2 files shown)


scorphus (commented):
Tricky one. I'm afraid it's better not to add code for such specific case.

Cases like this are much easier to solve by other means — e.g. by periodically exporting that history from the database to the shell's history file.

StandingPadAnimations:
Problem is that (as far as I know) Atuin doesn't provide a way to export commands to shell history

StandingPadAnimations:
Ok turns out Atuin already exports to the shell history, I'll close this now

PR #1373, opened 2023-04-17
Add `cd_quotes` rule

FireFragment:
```
> cd Some folder
cd: string not in pwd: Some
> fuck
cd "Some folder" [enter/↑/↓/ctrl+c]
```

This rule fixes forgotten quotes around directory containing spaces names when using `cd`.

The rule has priority 900 to override the `cd_mkdir` rule.


djh82 (commented):
Would be good to see some unit tests for this!

Issue #912, opened 2019-05-08
[feature request] Support nixos command-not-found


anka-213:
Nixos has a command-not-found command that outputs the messages in the following format

```
$ xmonad
The program ‘xmonad’ is currently not installed. You can install it by typing:
  nix-env -iA nixos.xmonad-with-packages
$ fuck
No fucks given

$ emacs
The program ‘emacs’ is currently not installed. It is provided by
several packages. You can install it by typing one of the following:
  nix-env -iA nixos.emacs
  nix-env -iA nixos.emacs25
  nix-env -iA nixos.emacs25-nox
  nix-env -iA nixos.emacs26-nox
```

The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    The Fuck 3.28 using Python 3.7.3 and Fish Shell 3.0.2

Your system (Debian 7, ArchLinux, Windows, etc.):

    nixos-19.03

How to reproduce the bug:

    Try to run a non-installed program on nixos
    command-not-found will automatically suggest how to install it
    run fuck
    thefuck doesn't give a fuck

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):

```
DEBUG: Run with settings: {'alter_history': True,
 'debug': True,
 'env': {'GIT_TRACE': '1', 'LANG': 'C', 'LC_ALL': 'C'},
 'exclude_rules': [],
 'history_limit': None,
 'instant_mode': False,
 'no_colors': False,
 'num_close_matches': 3,
 'priority': {},
 'repeat': False,
 'require_confirmation': True,
 'rules': [<const: All rules enabled>],
 'slow_commands': ['lein', 'react-native', 'gradle', './gradlew', 'vagrant'],
 'user_dir': PosixPath('/home/nixos-hjulle/.config/thefuck'),
 'wait_command': 3,
 'wait_slow_command': 15}
DEBUG: Received output: /bin/sh: xmonad: command not found

[...]

PR #1393, opened 2023-07-28
feat: new rule for `nix-shell`

thenbe:
Implementation is similar to the one explained in https://github.com/nvbn/thefuck/issues/912#issue-441679613.

In a nutshell, it tries to wrap the user's failed command in a nix-shell call.

```
$ ponysay moo
The program 'ponysay' is not in your PATH. You can make it available in an
ephemeral shell by typing:
  nix-shell -p ponysay

$ fuck
nix-shell -p ponysay --run "ponysay moo" [enter/↑/↓/ctrl+c]
```

Further info on nix-shell: https://thiagowfx.github.io/2022/02/nix-shell-in-a-nutshell/#hello-world-classic

### DIFF
--- a/thefuck/rules/nix_shell.py
+++ b/thefuck/rules/nix_shell.py
@@ -8,19 +8,11 @@ enabled_by_default = nix_available
 priority = 999
 
 
-nixpkgs_name = ""
-
-
 def get_nixpkgs_name(bin):
     """
     Returns the name of the Nix package that provides the given binary. It uses the
     `command-not-found` binary to do so, which is how nix-shell generates it's own suggestions.
     """
-
-    # Avoid getting the nixpkgs_name twice
-    global nixpkgs_name
-    if nixpkgs_name:
-        return nixpkgs_name
 
     result = subprocess.run(
         ["command-not-found", bin], stderr=subprocess.PIPE, universal_newlines=True
--- a/thefuck/rules/nix_shell.py
+++ b/thefuck/rules/nix_shell.py
@@ -1,0 +1,48 @@
+from thefuck.specific.nix import nix_available
+import subprocess
+
+enabled_by_default = nix_available
+
+# Set the priority just ahead of `fix_file` rule, which can generate low quality matches due
+# to the sheer amount of paths in the nix store.
+priority = 999
+
+
+nixpkgs_name = ""
+
+
+def get_nixpkgs_name(bin):
+    """
+    Returns the name of the Nix package that provides the given binary. It uses the
+    `command-not-found` binary to do so, which is how nix-shell generates it's own suggestions.
+    """
+
+    # Avoid getting the nixpkgs_name twice
+    global nixpkgs_name
+    if nixpkgs_name:
+        return nixpkgs_name
+
+    result = subprocess.run(
+        ["command-not-found", bin], stderr=subprocess.PIPE, universal_newlines=True
+    )
+
+    # return early if package is not available through nix
+    if "nix-shell" not in result.stderr:
+        return ""
+
+    nixpkgs_name = result.stderr.split()[-1] if result.stderr.split() else ""
+    return nixpkgs_name
+
+
+def match(command):
+    bin = command.script_parts[0]
+    return (
+        "nix-shell" not in command.script
+        and "command not found" in command.output
+        and get_nixpkgs_name(bin)
+    )
+
+
+def get_new_command(command):
+    bin = command.script_parts[0]
+    return 'nix-shell -p {0} --run "{1}"'.format(get_nixpkgs_name(bin), command.script)
--- a/README.md
+++ b/README.md
@@ -367,6 +367,7 @@ The following rules are enabled by default on specific platforms only:
 * `brew_update_formula` &ndash; turns `brew update <formula>` into `brew upgrade <formula>`;
 * `dnf_no_such_command` &ndash; fixes mistyped DNF commands;
 * `nixos_cmd_not_found` &ndash; installs apps on NixOS;
+* `nix_shell` &ndash; re-runs your command in a `nix-shell`;
 * `pacman` &ndash; installs app with `pacman` if it is not installed (uses `yay`, `pikaur` or `yaourt` if available);
 * `pacman_invalid_option` &ndash; replaces lowercase `pacman` options with uppercase.
 * `pacman_not_found` &ndash; fixes package name with `pacman`, `yay`, `pikaur` or `yaourt`.
--- a/tests/rules/test_nix_shell.py
+++ b/tests/rules/test_nix_shell.py
@@ -1,6 +1,5 @@
 import pytest
 from thefuck.rules.nix_shell import get_nixpkgs_name, get_new_command
-from thefuck.rules import nix_shell
 from thefuck.types import Command
 from unittest.mock import patch
 
@@ -15,9 +14,8 @@ mocked_nixpkgs = {
     "command, output", [("lsof", "lsof"), ("xev", "xorg.xev"), ("foo", "")]
 )
 def test_get_nixpkgs_name(command, output):
-    """ Check that `get_nixpkgs_name` returns the correct name """
+    """Check that `get_nixpkgs_name` returns the correct name"""
 
-    nix_shell.nixpkgs_name = ""
     with patch("subprocess.run") as mocked_run:
         instance = mocked_run.return_value
         instance.stderr = mocked_nixpkgs[command]
@@ -25,15 +23,18 @@ def test_get_nixpkgs_name(command, output):
 
 
 # check that flags and params are preserved for the new command
-@pytest.mark.parametrize('command_script, new_command', [
-    ('lsof -i :3000', 'nix-shell -p lsof --run "lsof -i :3000"'),
-    ('xev', 'nix-shell -p xorg.xev --run "xev"')])
+@pytest.mark.parametrize(
+    "command_script, new_command",
+    [
+        ("lsof -i :3000", 'nix-shell -p lsof --run "lsof -i :3000"'),
+        ("xev", 'nix-shell -p xorg.xev --run "xev"'),
+    ],
+)
 def test_get_new_command(command_script, new_command):
-    """ Check that flags and params are preserved in the new command """
+    """Check that flags and params are preserved in the new command"""
 
-    nix_shell.nixpkgs_name = ""
-    command = Command(command_script, '')
-    with patch('subprocess.run') as mocked_run:
+    command = Command(command_script, "")
+    with patch("subprocess.run") as mocked_run:
         instance = mocked_run.return_value
         instance.stderr = mocked_nixpkgs[command.script_parts[0]]
         assert get_new_command(command) == new_command
--- a/tests/rules/test_nix_shell.py
+++ b/tests/rules/test_nix_shell.py
@@ -1,0 +1,39 @@
+import pytest
+from thefuck.rules.nix_shell import get_nixpkgs_name, get_new_command
+from thefuck.rules import nix_shell
+from thefuck.types import Command
+from unittest.mock import patch
+
+mocked_nixpkgs = {
+    "lsof": "The program 'lsof' is not in your PATH. It is provided by several packages.\nYou can make it available in an ephemeral shell by typing one of the following:\n  nix-shell -p busybox\n  nix-shell -p lsof",
+    "xev": "The program 'xev' is not in your PATH. You can make it available in an ephemeral shell by typing:\n  nix-shell -p xorg.xev",
+    "foo": "foo: command not found",
+}
+
+
+@pytest.mark.parametrize(
+    "command, output", [("lsof", "lsof"), ("xev", "xorg.xev"), ("foo", "")]
+)
+def test_get_nixpkgs_name(command, output):
+    """ Check that `get_nixpkgs_name` returns the correct name """
+
+    nix_shell.nixpkgs_name = ""
+    with patch("subprocess.run") as mocked_run:
+        instance = mocked_run.return_value
+        instance.stderr = mocked_nixpkgs[command]
+        assert get_nixpkgs_name(command) == output
+
+
+# check that flags and params are preserved for the new command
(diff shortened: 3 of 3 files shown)


thefuck/rules/nix_shell.py
```diff
@@ -0,0 +1,40 @@
...
+
+    nixpkgs_name = result.stderr.split()[-1] if result.stderr.split() else ""
+    return nixpkgs_name
+
+
+def match(command):
+    bin = command.script_parts[0]
+    return (
+        "nix-shell" not in command.script
+        and "command not found" in command.output
```
scorphus, line 33: In your example:

```
$ ponysay moo
The program 'ponysay' is not in your PATH. You can make it available in an
ephemeral shell by typing:
  nix-shell -p ponysay

$ fuck
nix-shell -p ponysay --run "ponysay moo" [enter/↑/↓/ctrl+c]
```

`command not found` is not part of the output. Didn't you mean:
```suggestion
        and "command not found" not in command.output
```

scorphus (commented):
Thanks for contributing!

Please consider my comment below.

thefuck/rules/nix_shell.py
thenbe: > In your example, command not found is not part of the output

For commands that could be made available through nix:

1. Indeed, the visible output in the terminal does *not* contain the text "command not found".

```bash
$ spt
The program 'spt' is not in your PATH. You can make it available in an
ephemeral shell by typing:
  nix-shell -p spotify-tui
```

1. But when we log the value of `command` to file, we do see "command not found":

```
Command(script=spt, output=/nix/store/p6dlr3skfhxpyphipg2bqnj52999banh-bash-5.2-p15/bin/sh: line 1: spt: command not found)
```

I'm not sure exactly why they're different, but it works 🤷🏼‍♂️.

Admittedly, it could be simpler if we could do:

```diff
- and "command not found" in command.output
+ and command.exitcode is 127
```

But I couldn't find a way to access the _numeric_ exit code (127) from within the `match` function. Only the human [error message](https://tldp.org/LDP/abs/html/exitcodes.html) (command not found) is available it seems.

Finally, I've annotated the match conditions to make it clear what my intentions are for each.

scorphus: Ah, yes. That's because the command is executed through the shell — `/bin/sh`. Would you please run thefuck in debug mode and look for the line that says `DEBUG: Received output:`? e.g.:

```shell
THEFUCK_DEBUG=true thefuck spt
```

Among the debug messages there will be one with the output generated by the shell. Could you then please update the output values in `mocked_nixpkgs`?

Also, how about writing [tests for `match`](https://github.com/nvbn/thefuck/blob/62e0767c5069aeee176b0fe3459068b7703aaa26/tests/rules/test_brew_update_formula.py#L10-L19)? 😊

thenbe: > tests for match

done

I've cleaned up the PR a fair bit. As for the mock output, I've indicated which command I've used to get them so their purpose should be more clear now. If you notice that I've missed anything let me know. Thank you.

kdb424:
Been loving this. Hope it gets reviewed and upstreamed for all. Thanks for making this!

wesleycoder:
I've just discovered `thefuck` and I can't imagine how I didn't stumble into it earlier.

This PR would fit like a glove for me that just switched to nixos and haven't yet grown the muscle memory of typing the `nix-shell` whenever my command fails.

KiaraGrouwstra:
i would love to see this merged as well

KiaraGrouwstra:
looks like you [can use this already](https://github.com/NixOS/nixpkgs/compare/master...KiaraGrouwstra:nixpkgs:thefuck-nix-shell) using e.g. an overlay, altho i had a bit of trouble getting it to work out of the box.
specifically, without adding `doCheck = false;`, i would run into this error:

```
error: builder for '/nix/store/rl44gb6qd4x2myclj9i8cpkfrvw6ysqa-thefuck-3.32.drv' failed with exit code 2;
       last 10 log lines:
       > thefuck/system/unix.py:6
       >   /build/source/thefuck/system/unix.py:6: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
       >     from distutils.spawn import find_executable
       >
       > -- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
       > =========================== short test summary info ============================
       > ERROR  - ModuleNotFoundError: No module named 'pytest_docker_pexpect'
       > !!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
       > ========================= 1 warning, 1 error in 0.09s ==========================
       > /nix/store/bknngadwym46j65qs14ic2w79rpav888-stdenv-linux/setup: line 1582: pop_var_context: head of shell_variables not a function context
```

i had tried removing the added test, altho that appeared not to resolve the issue.

KiaraGrouwstra:
it would seem cool to similarly get an approach using `nix run`, i.e. go from suggesting `nix-shell -p ponysay --run "ponysay moo"` to `nix run nixpkgs#ponysay -- moo` - this might eventually help extend beyond just `nixpkgs`.

edit: https://github.com/KiaraGrouwstra/thefuck/commit/81d6786c80b86f2cc80b3ea90adc214df8266643

thenbe:
I've been using a custom rule that supports the new [unified CLI](https://zero-to-nix.com/concepts/nix#unified-cli) for a while, and was planning on opening a PR once this one has been merged (I hesitate to update this current PR as it's already tested and ready to be merged). I don't know if that will happen soon, so in the meantime I've pushed the changes to [this](https://github.com/thenbe/thefuck/tree/nix-shell-new) new branch instead, which [builds](https://github.com/thenbe/thefuck/compare/nix-shell...thenbe:thefuck:nix-shell-new) on this here PR. You can use the updated rule by adding it as a [custom rule](https://github.com/nvbn/thefuck?tab=readme-ov-file#creating-your-own-rules) to your config.

In the new rule, three variants are suggested. Assuming I run `cowsay hello world`, I am presented with the following:

1. `nix run nixpkgs#cowsay -- hello world`: This runs my command in a non-interactive shell. Uses the nix unified CLI.
1. `nix shell nixpkgs#cowsay`: This enters an interactive shell with `cowsay` available, but does not run any command. This is useful if you'd rather run the command yourself after entering the shell because your command requires delicate massaging (e.g. running it with `sudo`, prefixing it with environment variable, juggling quote variants, etc).
1. `nix-shell -p cowsay --run "cowsay hello world"`. This runs my command in a non-interactive shell. Uses the nix original CLI.
1. `nix shell nixpkgs#cowsay --command cowsay hello world`: Very similar to the first one so I've personally disabled this one.

### Thoughts on future updates:



- It'd be nice if there was a variant that runs my command and then _keeps me_ in the shell.
  - For the original CLI, we [can](https://nix.dev/manual/nix/2.19/command-ref/nix-shell#options) add a `--command "echo hello; return"` to our `nix-shell` invocation.
  - For the unified CLI: not sure yet, we might need to do something like this example from the [docs](https://nix.dev/manual/nix/2.19/command-ref/new-cli/nix3-shell): ` nix shell nixpkgs#gnumake --command sh -c "cd src && make"`
- We should expose a couple of flags for users to configure this.
  - `disable_unified_cli` (boolean)
  - `disable_original_cli` (boolean)
- As far as I can tell, the `command-not-found` db doesn't really play nice if you use flakes to configure your system and might return stale results (unless you update it manually?). [`nix-index`](https://github.com/nix-community/nix-index) seems to be the go-to alternative. It'd be great if we could optionally use that instead (perhaps behind a flag `enable_nix_index` for users who have installed `nix-index` (`programs.nix-index.enable = true;` in home manager).

KiaraGrouwstra:
@thenbe i agree integrating with `nix-index`'s `command-not-found` replacement seems cool, as a flake user.
i kinda wish we could have `command-not-found` (and this `thefuck` integration) [extend to flake inputs beyond nixpkgs](https://github.com/nix-community/nix-index/issues/244) as well, such as to packages from NUR for example. preferably this should be dynamic based on your inputs rather than hardcoded to specific ones like nixpkgs, or NUR for that matter.
i'll admit i haven't really figured out how that might work tho.

KiaraGrouwstra:
just tried these with a command like `program_i_have | program_i_dont_have`, seems that may complicate the suggestions a bit

thenbe:
I'm not sure if `thefuck` can handle piping.

If I make a typo `git statis` it will correct me to `git status`. But if I do `echo hello | git statis` it does not correct my typo. `thefuck` seems to work mostly on single commands AFAICT.

KiaraGrouwstra:
@thenbe hm, i'm not sure.

```
fortune | cowsay
The program 'cowsay' is not in your PATH. It is provided by several packages.
You can make it available in an ephemeral shell by typing one of the following:
  nix-shell -p cowsay
  nix-shell -p neo-cowsay
$ fuck
nix run nixpkgs#fortune | cowsay
```

feels like it knows about the whole command given it's reproducing it?

KiaraGrouwstra:
another common nix thing we might be able to address from `thefuck` would be errors about packages being unfree

edit: https://github.com/KiaraGrouwstra/thefuck/commit/16d838bf6f63117b161a2f1e6572e06108b007eb

KiaraGrouwstra:
@thenbe what was the argument to favor `nix run` over `nix shell` again? i guess the latter seems a bit more generic in case of handling non-standard binaries at least

thenbe:
If I'm only looking to execute a program (and don't need to be dropped into a shell) then I prefer `nix run` over `nix shell` as the [documentation](https://nix.dev/manual/nix/2.19/command-ref/new-cli/nix3-run) suggests `nix run` specifically for this use case.

I also recall `nix run` being more performant (perhaps because we forego the overhead of launching a shell?). This last point is not derived from benchmarks, only anecdotal evidence.

> i guess the latter seems a bit more generic in case of handling non-standard binaries at least

I've added this variant (the 4th one in my [previous post](https://github.com/nvbn/thefuck/pull/1393#issuecomment-1961487094)), but disabled it after a while when I realized that I never reach for it. Do you find that you still need it over `nix run` (the 1st variant in my previous post)?

thenbe:
> another common nix thing we might be able to address from `thefuck` would be errors about packages being unfree
> 
> edit: [KiaraGrouwstra@16d838b](https://github.com/KiaraGrouwstra/thefuck/commit/16d838bf6f63117b161a2f1e6572e06108b007eb)

This would be useful. Does it still complain about the `--impure` flag? Or do you use a workaround for that?

KiaraGrouwstra:
i've been using `thefuck` mostly thru its [`zsh` plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/thefuck), which just gets you the top suggestion. i found that failed for me for e.g. `poppler_utils`, which bundles multiple binaries.
to be fair tho, i'm not sure that accounts for a large portion of its invocations, so maybe it could make sense to just actually type out `fuck` in those cases.

what was the `--impure` error? i'm not sure i'd run into that.

by the way, had you managed to also package your branch for nix? considering i seemed to need that `doCheck = false;` to get our branches to build thru nix.

thenbe:
I just have it aliased to `f` for extra convenience.

I opted not to package it for nix separately since `fuck` already exposes a method for easily adding custom rules. Instead, I placed the rule in `~/mydotfiles/thefuck/rules/nix-shell.py` then told home-manager to symlink it to the appropriate place in `.config`:

```nix
# home.nix
home.file.".config/thefuck/rules/nix-shell.py".source = config.lib.file.mkOutOfStoreSymlink "${config.home.homeDirectory}/mydotfiles/thefuck/rules/nix-shell.py";
```

This way I don't need to rebuild every time I tweak the rule.

> what was the --impure error?

The unified CLI commands (`nix shell`, `nix run`, etc) will not acknowledge environment variables unless the `--impure` flag is used.

<details>
  <summary> output </summary>

```
$ NIXPKGS_ALLOW_UNFREE=1 nix shell nixpkgs#github-copilot-cli

error:
       … in the condition of the assert statement

         at /nix/store/xwc3zfc544jg6zhr0wi6k8253s7mwlhi-source/lib/customisation.nix:267:17:

          266|     in commonAttrs // {
          267|       drvPath = assert condition; drv.drvPath;
             |                 ^
          268|       outPath = assert condition; drv.outPath;

       … while evaluating the attribute 'handled'

         at /nix/store/xwc3zfc544jg6zhr0wi6k8253s7mwlhi-source/pkgs/stdenv/generic/check-meta.nix:490:7:

          489|       # or, alternatively, just output a warning message.
          490|       handled =
             |       ^
          491|         (

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: Package ‘github-copilot-cli-0.1.36’ in /nix/store/xwc3zfc544jg6zhr0wi6k8253s7mwlhi-source/pkgs/tools/misc/github-copilot-cli/default.nix:21 has
 an unfree license (‘unfree’), refusing to evaluate.

       a) To temporarily allow unfree packages, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_UNFREE=1

          Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
                then pass `--impure` in order to allow use of environment variables.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowUnfree = true; }
       in configuration.nix to override this.

       Alternatively you can configure a predicate to allow specific packages:
         { nixpkgs.config.allowUnfreePredicate = pkg: builtins.elem (lib.getName pkg) [
             "github-copilot-cli-0.1.36"
           ];
         }

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowUnfree = true; }
       to ~/.config/nixpkgs/config.nix.


```
```bash
# it wants this instead:
$ NIXPKGS_ALLOW_UNFREE=1 nix shell nixpkgs#github-copilot-cli --impure
```

</details>

KiaraGrouwstra:
aah i see! i'd yet to take that into account. 🙈

specifying the rules rather than doing overlays makes sense - thanks!

PR #1437, opened 2024-02-28
Update tests: mock is now part of unittest

aylen384:


dvzrv:
It would be great if this could be merged and released, as we need this for removing python-mock on Arch Linux! :cat:

PR #1447, opened 2024-05-15
Change USER_COMMAND_MARK

olavfla:
Change USER_COMMAND_MARK to include escape characters so that the non-printing characters are not counted towards prompt length.


olavfla:
Without this, experimental instant mode has an issue where line breaks occur 10 characters too early, making the terminal a bit wonky. Only tested with one terminal though

olavfla:
On a slightly related note, in def instant_mode_alias(self, alias_name) in shells/bash.py, is it really necessary with the backspaces in `mark = USER_COMMAND_MARK + '\b' * len(USER_COMMAND_MARK)`? In my experience, it works just fine without them, and fucks up if anything appends something to the start of PS1 (such as enabling a python venv).

Issue #1428, opened 2024-01-11
Shell slow to start with `eval "$(thefuck --alias)"`, workaround is a lazy loading trick


mooreye:
**The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):**

`The Fuck 3.32 using Python 3.12.1 and Bash 5.2.21(1)-release`

**Your system (Debian 7, ArchLinux, Windows, etc.):**

`Fedora 39`

**How to reproduce the bug:**

Put `eval $(thefuck --alias)` in `~/.bashrc` and observe that the shell is noticeably slower to start than without it.

Time without:

```
$ time bash -ic :
                                                                                                                                            
real	0m0.055s
user	0m0.037s
sys	0m0.019s
```

Time with:

```
$ time bash -ic :
                                                                                                                                            
real	0m0.263s
user	0m0.219s
sys	0m0.043s
```

**Anything else you think is relevant:**

If there is nothing that can be done to fix this, the following lazy loading trick works for me:

`alias fuck='if ! declare -f fuck &>/dev/null; then eval -- "$(thefuck -a)"; fi && fuck'`

Maybe this should be added to the docs somehow?

pallaswept:
Came to report this same issue. For some reason, `thefuck --alias` takes ~100ms to return, on a 5900X... Similar functions (zoxide init, starship init, etc) take ~2ms here.

```
time thefuck --alias

            function fuck () {
                TF_PYTHONIOENCODING=$PYTHONIOENCODING;
                export TF_SHELL=bash;
                export TF_ALIAS=fuck;
                export TF_SHELL_ALIASES=$(alias);
                export TF_HISTORY=$(fc -ln -10);
                export PYTHONIOENCODING=utf-8;
                TF_CMD=$(
                    thefuck THEFUCK_ARGUMENT_PLACEHOLDER "$@"
                ) && eval "$TF_CMD";
                unset TF_HISTORY;
                export PYTHONIOENCODING=$TF_PYTHONIOENCODING;
                history -s $TF_CMD;
            }
        

real    0m0.097s
user    0m0.085s
sys     0m0.012s
[elapsed: 97ms (CPU 99.0%)] time thefuck --alias
```

My workaround rather than lazy loading was simply to use the output of the command directly in my bashrc. I'll need to re-do that if the output changes and `thefuck` breaks in future, but the performance is better.

PR #1404, opened 2023-10-08
replace distutils for python 3.12

branchvincent:
Python 3.12 has removed `distutils`: https://docs.python.org/3.12/whatsnew/3.12.html#removed

PR #1459, opened 2024-08-23
Reduce Startup Execution Time

cyuria:
Reduces the impact of #1428

Nearly half of the startup time of running `thefuck -a` is caused by the removed line of code in `thefuck/system/unix.py`.  This line of code imports `find_executable` from `distutils`, however this import takes on the order of 40% of the total startup time. This means my personal startup time has been reduced from about 400ms to 250ms or so as seen in the below image.

image

I've attempted to find further ways of reducing this time, but bar from fully isolating the entire chain of functions, it looks like this isn't particularly feasible.

One thing I haven't tried is to use lazy importing of some kind. This could yield serious performance improvements, as most of the delays are caused by `import` statements which, for the most part, aren't used by the `--alias` code path, however are not possible to easily move or remove due to their numerous quantity and interlinked dependencies in just about every source file. Transitioning to lazy loading would likely take a fair bit of effort to make sure everything still works and there is still no guarantee of better performance, however it is a possible avenue to continue down.

### DIFF
--- a/thefuck/system/unix.py
+++ b/thefuck/system/unix.py
@@ -3,7 +3,6 @@ import sys
 import tty
 import termios
 import colorama
-from distutils.spawn import find_executable
 from .. import const
 
 init_output = colorama.init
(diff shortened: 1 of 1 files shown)


cyuria:
It looks like this might conflict with #1404

I'd prefer that get merged over this PR, however I will leave this open until one of these two does get merged.

PR #1465, opened 2024-09-08
Replace deprecated imp module with importlib for Python compatibility

teja-pola:
Replaced all usages of the deprecated imp module with importlib to ensure compatibility with Python 3.12 and beyond.

### DIFF
--- a/thefuck/conf.py
+++ b/thefuck/conf.py
@@ -14,7 +14,8 @@ try:
         module_spec.loader.exec_module(module)
         return module
 except ImportError:
-    from imp import load_source
+    # This block is no longer needed if importlib.util is always available.
+    pass
 
 
 class Settings(dict):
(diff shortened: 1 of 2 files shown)


~/.config/thefuck/settings.py
cyuria, line 1: I'm not sure exactly what is going on here

Edit: It looks like somehow your settings have made it into the PR

thefuck/conf.py
cyuria, line 1: If you aren't doing anything on error, surely it would just be better to remove the `try` entirely. If `load_source` isn't available, it will fail later on with the instantiation of `Settings` at the bottom anyway, leading to a more confusing error message.

cyuria (commented):
Regardless it seems to me like this repo isn't really maintained at the moment. There are a number of open PRs which are ready to be merged as far as I can tell, but haven't been. I have opened an issue #1466 with regards to the status of the project, so hopefully someone will pick up the maintenance and your PR can be properly sorted out.

Issue #1460, opened 2024-08-27
I get a error when i run fuck command in windows, installed fuck with pip


itzmesonic:
The output of `thefuck --version` (something like `The Fuck 3.1 using Python
3.5.0 and Bash 4.4.12(1)-release`):

    Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Scripts\thefuck.exe\__main__.py", line 4, in <module>
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Lib\site-packages\thefuck\entrypoints\main.py", line 8, in <module>
    from .. import logs  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Lib\site-packages\thefuck\logs.py", line 8, in <module>
    from .conf import settings
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Lib\site-packages\thefuck\conf.py", line 1, in <module>
    from imp import load_source
ModuleNotFoundError: No module named 'imp'

Your system (Debian 7, ArchLinux, Windows, etc.):

    Windows 11

How to reproduce the bug:

    No idea

The output of The Fuck with `THEFUCK_DEBUG=true` exported (typically execute `export THEFUCK_DEBUG=true` in your shell before The Fuck):
```
   Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Scripts\thefuck.exe\__main__.py", line 4, in <module>
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Lib\site-packages\thefuck\entrypoints\main.py", line 8, in <module>
    from .. import logs  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Lib\site-packages\thefuck\logs.py", line 8, in <module>
    from .conf import settings
  File "C:\Users\mardin it\AppData\Local\Programs\Python\Python312\Lib\site-packages\thefuck\conf.py", line 1, in <module>
    from imp import load_source
ModuleNotFoundError: No module named 'imp'
[...]

PR #1473, opened 2024-09-21
Bugfix

Coder-Harshit:
REFERENCE: #1460 

CHANGES:

- `shutil` module in use (instead of distutils ... which is gonna soon be removed)
- updated `requirements.txt` file
- removed `imp` module (due to deprection)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ pytest-benchmark
 pytest-docker-pexpect
 twine
 pathlib2
+importlib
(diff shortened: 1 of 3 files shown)


itzmesonic:
How do i build your repository version??
I wanna test it

Coder-Harshit:
> How do i build your repository version??
> I wanna test it

Would it be OK if I say I have no idea what you are talking about .... as per my inference from the main repo to develope or test new features we just install it via pip under a virtual environment...is that correct ?

itzmesonic:
> > How do i build your repository version??
> > I wanna test it
> 
> Would it be OK if I say I have no idea what you are talking about .... as per my inference from the main repo to develope or test new features we just install it via pip under a virtual environment...is that correct ?

I say
**how do i convert it to a setup.exe file??**

PR #1475, opened 2024-10-02
Fix compatibility with Python > 3.10

DJStompZone:
#### Problem

The `thefuck` project currently uses the `imp` module for loading source files dynamically, but this module has been deprecated since Python 3.4 and removed entirely in Python 3.10. As a result, users running Python 3.10+ encounter the following error:

```
ModuleNotFoundError: No module named 'imp'
```

#### Solution

This PR introduces a solution that maintains backward compatibility by replacing `imp` with `importlib.util` for Python 3.5+ while providing a fallback to `importlib.machinery.SourceFileLoader` for Python 3.3 and 3.4. For even older versions, it keeps the final fallback to `imp`, ensuring compatibility across a wide range of Python versions.

#### Changes

- Replaced the `imp` module import with the following structure:
  - Use `importlib.util` for Python 3.5+.
  - Use `importlib.machinery.SourceFileLoader` for Python 3.3 and 3.4.
  - Retain the `imp` module for Python versions below 3.3.
- Ensured the changes do not affect existing functionality on older Python versions.
  
#### Impact

- Python 3.10+ users should no longer encounter the `ModuleNotFoundError`.
- Existing users on older Python versions should see no changes in behavior.
- Windows should now be fully compatible

#### Resolved Issues

This PR closes the following issues:
- Fixes #1460, Fixes #1464, Fixes #1467, Fixes #1434, Fixes #1433, Fixes #1418, Fixes #1381 

I'm hoping that's all of them, sorry if I missed any.

PR #1487, opened 2024-12-26
♻️ refactor(unix): replace deprecated distutils.spawn with shutil.which

lujin3:
fix: ModuleNotFoundError: No module named 'distutils'
- remove deprecated distutils.spawn import
- add shutil import
- implement find_executable using shutil.which for better compatibility

### DIFF
--- a/thefuck/system/unix.py
+++ b/thefuck/system/unix.py
@@ -3,8 +3,12 @@ import sys
 import tty
 import termios
 import colorama
-from distutils.spawn import find_executable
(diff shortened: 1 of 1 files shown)


JamesConlan96:
Bump

swirle13:
@lujin3 Is this an improvement over #1404?

What versions of python does this work for? Does it do more than #1475 which fixes this for python > 3.10 and prevents import errors for old python versions as old as 3.3

PR #1516, opened 2025-07-10
Changes to induce compatablity with python 12+ 

RushilCodes:
Some minute changes to thefuck/conf.py and thefuck/types.py to make it compatible with python 12+(python 13) which does not support use of imp lib, i have used some code as a drop-in replacement, which will make it easier to check and extract into a file if so becomes a requirement

### DIFF
--- a/thefuck/conf.py
+++ b/thefuck/conf.py
@@ -1,20 +1,19 @@
+import importlib.util
+import sys
+import os
+
+def load_source(name, path):
+    spec = importlib.util.spec_from_file_location(name, path)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    spec.loader.exec_module(module)
+    return module
 import os
 import sys
 from warnings import warn
 from six import text_type
 from . import const
 from .system import Path
-
-try:
-    import importlib.util
-
-    def load_source(name, pathname, _file=None):
-        module_spec = importlib.util.spec_from_file_location(name, pathname)
-        module = importlib.util.module_from_spec(module_spec)
-        module_spec.loader.exec_module(module)
-        return module
-except ImportError:
-    from imp import load_source
 
 
 class Settings(dict):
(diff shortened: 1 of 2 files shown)


OrangeDoro (commented):
Hi! I'm a grad student working on a research project about using large language models to automate code review. Based on your commit 7cb8b8e8c27029e9867b1a2a4b33f1f631094193 and the changes in thefuck/types.py, my tool generated this comment:
1. Add error handling to manage cases where the module cannot be executed.
2. 2. Implement checks to prevent path traversal attacks.
3. 3. Validate and sanitize the `path` input to ensure it points to a trusted source.
4. 4. Ensure that the `path` parameter is valid or that the file exists at the specified location before attempting to load it.
5. 5. Validate the `name` parameter to ensure it is a non-empty string and does not conflict with existing module names.
6. 6. Wrap the `spec.loader.exec_module(module)` call in a try-except block to handle potential exceptions gracefully.
7. 7. Check for existing modules in `sys.modules` before overwriting them.
8. 8. Ensure that all references to the old `load_source` function are updated if the intention is to replace it.
9. 9. If the original `load_source` function is still needed, keep the import or ensure the new definition is used consistently.
10. 10. Add unit tests to verify the behavior of `load_source`, covering successful loading, handling of invalid paths, and exceptions raised during the loading process.
11. 11. Consider moving `load_source` back to the `.conf` module or another shared utility module to promote reuse and reduce code duplication.
12. 12. Add type hints to the `load_source` function to enhance readability and maintainability.

As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:


1. Does this comment provide suggestions from a dimension you hadn’t considered?

2. Do you find this comment helpful?

Thanks a lot for your time and feedback! And sorry again if this message is a bother.

RushilCodes:
the comments have no relation to my commit, they are related to the project but not my commits

PR #1564, opened 2026-03-30
fix: use shell-agnostic 'and' operator in sudo rule instead of hardcoded '&&'

### LINKED ISSUE #1229: sudo.py assumes && for shell 'and'
The sudo.py rule assumes that '&&' is every shells 'and' operator. This is not the case for fish shell v2.

juliosuas:
on fish this rule doesn't fire, because the code is using `&&` literally instead of asking the shell what the 'and' operator is.

other rules (pacman, git_stash, rails_migrations_pending...) already go through `shell.and_()` for this exact reason. switched sudo to do the same.

while at it, `match()` was checking `command.script_parts` which is already tokenized. if the operator is multi-token (`; and `) it gets split across parts and the check misses. changed to check against `command.script` directly.

fixes #1229.


juliosuas:
Context for reviewers: the `sudo` rule was checking for `'&&'` as a hardcoded string, but Fish shell v2 uses `'; and '` instead. The fix uses `shell.and_()` — the same approach already used in `pacman.py`, `git_stash.py`, and `rails_migrations_pending.py`. Also fixed the match check to look at `command.script` (full string) rather than `command.script_parts` (list), since the operator may appear across multiple tokens.

juliosuas:
Closing this stale PR to keep the queue lean. The shell-agnostic idea is still reasonable, but no need to leave old review burden open.

